PR #2, opened 2017-11-01
Add Travis-CI config and scripts

matyasselmeci:
In order to keep one of the tests from timing Travis-CI out, I had to decrease the timeout of the test.
Tests need to run as non-root or various authentication things will fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 sudo: required
 env:
     matrix:
-    - OS_VERSION=6
-    - OS_VERSION=7
+    - OS_VERSION=6 COMPONENTS=udt,myproxy
+    - OS_VERSION=7 COMPONENTS=udt,myproxy
+    - OS_VERSION=6 COMPONENTS=gram5
+    - OS_VERSION=7 COMPONENTS=gram5
 
 services:
     - docker
@@ -13,5 +15,5 @@ before_install:
     - sudo docker pull centos:centos${OS_VERSION}
 
 script:
-- travis-ci/setup_tests.sh
+- travis-ci/setup_tests.sh ${COMPONENTS}
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,5 @@ before_install:
     - sudo docker pull centos:centos${OS_VERSION}
 
 script:
-- travis-ci/setup_tests.sh ${COMPONENTS}
+- travis-ci/setup_tests.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: required
 env:
     matrix:
-    - OS_VERSION=6 COMPONENTS=udt,myproxy
-    - OS_VERSION=7 COMPONENTS=udt,myproxy
+    - OS_VERSION=6 COMPONENTS=udt,myproxy,ssh
+    - OS_VERSION=7 COMPONENTS=udt,myproxy,ssh
     - OS_VERSION=6 COMPONENTS=gram5
     - OS_VERSION=7 COMPONENTS=gram5
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+sudo: required
+env:
+    matrix:
+    - OS_VERSION=6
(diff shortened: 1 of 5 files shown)


xio/src/test/timeout_test.c
```diff
@@ -25,7 +25,7 @@ static globus_bool_t                    globus_l_closed = GLOBUS_FALSE;
 
 static globus_bool_t                    globus_l_timeout = GLOBUS_TRUE;
 
-#define USEC_THRESHHOLD  500000
+#define USEC_THRESHHOLD  50000
```
brianhlin: Alternatively, we can extend the travis timeout with [travis_wait](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received)

bbockelm: this is 500ms versus 50ms, right?  Why are we so sensitive to this?

@matyasselmeci - how long does this take to run without this change?

matyasselmeci:
Interesting. I reran the test with the original timeout and it didn't seem to be a problem this time. I'll remove the change.

matyasselmeci:
Fixed. Anything else?

brianhlin:
Can we enable this repo in Travis-CI? It would be nice to have results in the PR

matyasselmeci:
I don't have access to do that (though I just sent a request via Travis-CI).

bbockelm:
I think this one is ready to go.

However, I'm going to hold off on pressing the merge button until we get some more of the organizational stuff fixed up (hopefully Monday?).

Next up - source tarball, binary tarball, and RPMs?  I'd like to figure out how we post the 58 individual source tarballs for each component for Mattias Ellert on releases.

matyasselmeci:
I don't think MyProxy is getting built. Its `configure` script fails if run because Globus dependencies aren't installed (bootstrapping issue, eh?).

matyasselmeci:
OK, I now build and test myproxy, udt, and optional gram5 components. The gram5 build has been separated out into its own set of Travis tasks, because otherwise we hit the Travis log size limit (which they give no workaround for).

The only thing I don't build is gridftp-hdfs, because that requires cloudera's repos to be available so I can install hadoop, and I don't want to do that unless we really have to. gridftp-hdfs has no unit tests, and is not packaged in EPEL anyway.

The Travis build is here: https://travis-ci.org/matyasselmeci/gct/builds/301981256. (There is a build 52 as well, but it's the same as build 51 but with some commits squashed.)

@bbockelm, could you please re-review?

matyasselmeci:
I am working on an alternate implementation that uses the Travis default Ubuntu environment instead of Docker containers inside VMs. I think we can get away with doing that instead of what's in this PR because this Travis build is not testing packaging, only compilation. The build is faster by ~10 minutes

If you think that's a good idea, I will submit an alternate PR that we should merge instead of this.

matyasselmeci:
The outside-the-box experiment isn't going that well (or at least, I can't get many of the GRAM tests to pass). I'd say it's better to merge this PR as-is instead of spending more time on it. @bbockelm or @brianhlin, could you review and merge? And approve my Travis access to the gridcf organization?

bbockelm:
Still looks good to me -- what's needed on Travis?

matyasselmeci:
OK. Added a few tweaks and squashed some commits: https://travis-ci.org/matyasselmeci/gct/builds/303135017

.travis.yml
```diff
@@ -0,0 +1,27 @@
+sudo: required
+env:
+    matrix:
+    - IMAGE=centos:centos6 COMPONENTS=udt,myproxy,ssh
+    - IMAGE=centos:centos7 COMPONENTS=udt,myproxy,ssh
+#    - IMAGE=fedora:25      COMPONENTS=udt,myproxy,ssh
+#    - IMAGE=fedora:26      COMPONENTS=udt,myproxy,ssh
+#    - IMAGE=fedora:27      COMPONENTS=udt,myproxy,ssh
```
brianhlin: Could we either remove these comments or get explanatory text for why they're commented out?

PR #3, opened 2017-11-21
Merge changes from globus/globus-toolkit repo made since our fork

matyasselmeci:


bbockelm:
I see a few version numbers have been bumped upstream - is there anything we need to consider here?

msalle:
Wouldn't it make sense, at least for the time being, to track directly the globus-toolkit as an upstream branch?

matyasselmeci:
I'd be fine with that. It would give us a use for `globus_6_branch`.

matyasselmeci:
OK, merging. For `osg-bot`, I guess I'd like something that pulls their globus_6_branch into ours, then does a force push. And in the future, we'll make pull requests into master from our globus_6_branch.

PR #6, opened 2017-11-29
Make source tarballs in Travis

matyasselmeci:
Add a script that builds a full set of source tarballs for the GCT components (except gridftp-hdfs). Can be run standalone easily, but added config to run via Travis: https://travis-ci.org/matyasselmeci/gct/builds/308697470

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,23 @@
 sudo: required
 env:
     matrix:
-    - IMAGE=centos:centos6 COMPONENTS=udt,myproxy,ssh
-    - IMAGE=centos:centos7 COMPONENTS=udt,myproxy,ssh
+    - IMAGE=centos:centos6 TASK=tests COMPONENTS=udt,myproxy,ssh
+    - IMAGE=centos:centos7 TASK=tests COMPONENTS=udt,myproxy,ssh
 #   Fedora builds are implemented but are disabled until we decide we
 #   need them; Travis limits our number of concurrent jobs so we don't
 #   want more than are necessary.
-#    - IMAGE=fedora:25      COMPONENTS=udt,myproxy,ssh
-#    - IMAGE=fedora:26      COMPONENTS=udt,myproxy,ssh
-#    - IMAGE=fedora:27      COMPONENTS=udt,myproxy,ssh
-    - IMAGE=centos:centos6 COMPONENTS=gram5
-    - IMAGE=centos:centos7 COMPONENTS=gram5
+#    - IMAGE=fedora:25     TASK=tests COMPONENTS=udt,myproxy,ssh
+#    - IMAGE=fedora:26     TASK=tests COMPONENTS=udt,myproxy,ssh
+#    - IMAGE=fedora:27     TASK=tests COMPONENTS=udt,myproxy,ssh
+    - IMAGE=centos:centos6 TASK=tests COMPONENTS=gram5
+    - IMAGE=centos:centos7 TASK=tests COMPONENTS=gram5
 #   The following gram tests failed for all fedora versions:
 #       nonblocking-register-test.pl, register-callback-test.pl, register-test.pl
-#    - IMAGE=fedora:25      COMPONENTS=gram5
-#    - IMAGE=fedora:26      COMPONENTS=gram5
-#    - IMAGE=fedora:27      COMPONENTS=gram5
+#    - IMAGE=fedora:25      TASK=tests COMPONENTS=gram5
+#    - IMAGE=fedora:26      TASK=tests COMPONENTS=gram5
+#    - IMAGE=fedora:27      TASK=tests COMPONENTS=gram5
+    - IMAGE=centos:centos7 TASK=tarballs COMPONENTS=ssh,gram5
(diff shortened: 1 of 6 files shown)


bbockelm:
Can you make this into a pipeline?  Only produce the tar balls if all the builds succeed?

fscheiner:
Not sure if this:

> from https://github.com/gridcf/gct/pull/2#issuecomment-341290336 by @bbockelm :
> [...]
> I'd like to figure out how we post the 58 individual source tarballs for each component for Mattias Ellert on releases.

...was already solved, but at least size seems to be no issue according to [GitHub](https://help.github.com/articles/distributing-large-binaries/) - **less than 2 GB per file** and **no limit for the total size of all files** in releases. Though I don't know, how this will look in a few years with a lot of releases and millions of downloads... ;-)

And it also seems to be possible for Travis CI to deploy to GitHub releases directly according to the [corresponding documentation](https://docs.travis-ci.com/user/deployment/releases/).

fscheiner:
@matyasselmeci :
Would it be possible to workaround the time constraints by using `travis_wait` as mentioned on https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received?

matyasselmeci:
No, travis_wait is for working around the 10-minutes-with-no-output timeout; the one I would run into is the 50-minutes-per-job timeout. (Also the 4 MB of logs per job limit.) I think with Build Stages, I can make the deploy stage happen only if the test stages succeed.

matyasselmeci:
Can we merge this and have deployment/uploading be a separate PR? I've been working on making RPMs, which depends on this.

brianhlin:
Why aren't we seeing CI builds getting kicked off?

fscheiner:
@brianhlin
I just had a look into [the hook settings for Travis CI](https://github.com/gridcf/gct/settings/hooks/18227087) (reached from [Integrations & services](https://github.com/gridcf/gct/settings/installations)) and both the fields for **User** and **Token** are currently empty. I did not change anything there, just had a look today. I assume these configs are needed for this hook to work, right?

So did GitHub perhaps forgot these configurations or did someone change the configuration there to try something out, maybe because of my https://github.com/gridcf/gct/pull/7#issuecomment-347790274 regarding Travis CI?

brianhlin:
@fscheiner I don't believe that's the case because it's enabled in the Travis-CI interface so the user/token shouldn't matter. The [Travis docs](https://docs.travis-ci.com/user/pull-requests/) have this to say:

> If a pull request isn’t built or doesn’t show up in Travis CI’s user interface, that usually means that it can’t be merged. We rely on the merge commit that GitHub transparently creates between the changes in the source branch and the upstream branch the pull request is sent against.

I don't see anything about mergeability with master. Maybe that's the issue?

matyasselmeci:
Yes, that is weird. I can't find a way in Travis to trigger it either, or change the settings on the repo to generate a new token or something.

Here is the most recent build of this branch on my fork:
https://travis-ci.org/matyasselmeci/gct/builds/310064911?utm_source=github_status&utm_medium=notification

brianhlin:
I think I figured it out: somehow the protected master branch config did not require the Travis-CI status check before merging

.travis.yml
```diff
@@ -1,21 +1,23 @@
...
+    - IMAGE=centos:centos7 TASK=tests COMPONENTS=gram5
 #   The following gram tests failed for all fedora versions:
 #       nonblocking-register-test.pl, register-callback-test.pl, register-test.pl
-#    - IMAGE=fedora:25      COMPONENTS=gram5
-#    - IMAGE=fedora:26      COMPONENTS=gram5
-#    - IMAGE=fedora:27      COMPONENTS=gram5
+#    - IMAGE=fedora:25      TASK=tests COMPONENTS=gram5
+#    - IMAGE=fedora:26      TASK=tests COMPONENTS=gram5
+#    - IMAGE=fedora:27      TASK=tests COMPONENTS=gram5
+    - IMAGE=centos:centos7 TASK=tarballs COMPONENTS=ssh,gram5
```
brianhlin: If I'm reading this right, we don't do anything with the `COMPONENTS` for the `tarballs` task, right?

matyasselmeci: We don't, but it does determine which RPMs get yum installed first

matyasselmeci:
It's been "waiting for status to be reported" for several hours now; Travis does not show that a build corresponding to this PR has ever been run.

brianhlin:
Hrm, all the Travis/Github settings look ok to me at this point. Could you push a dummy commit to see if that triggers a build? It looks like we started triggering Travis builds off this repo [15 days ago](https://travis-ci.org/gridcf/gct/requests) and your latest commit is from 19 days ago so it may just be as simple as that.

matyasselmeci:
Tried it; did not seem to work.

matyasselmeci:
Never mind, figured it out. I already had a PR of this branch against _my_ master, and I guess Travis didn't want to start another build of the exact same hash. I closed that PR, made a dummy commit, and the build started up for this now.

PR #8, opened 2017-12-06
Make RPMs in Travis

matyasselmeci:
Add make_rpms.sh script that can be run by hand or via Travis as the `rpms` task. Since building the RPMs requires building the tarballs first, the `tarballs` task is no longer run by itself.

The gsi_openssh RPM does not work with the tarball we create because it uses the official openssh tarball and applies some patches itself (the tarball we make is pre-patched). We plan to use the EPEL version anyway, so I dropped the spec file.


travis-ci/run_task_inside_docker.sh
```diff
@@ -0,0 +1,103 @@
...
+TASK=$2
+COMPONENTS=$3
+
+
+set -xe
+
+# Clean the yum cache
+yum clean all
+
+yum install -y -d0 patch
```
brianhlin: Is `patch` required for any task other than `rpms`?

travis-ci/run_task_inside_docker.sh
```diff
@@ -0,0 +1,103 @@
...
+        su builduser -c "/bin/bash -xe /gct/travis-ci/build_and_test.sh $IMAGE $COMPONENTS"
+        ;;
+    (tarballs)
+        cd /gct
+        su builduser -c "/bin/bash -xe /gct/travis-ci/make_source_tarballs.sh"
+        print_file_size_table Tarballs /gct/package-output
+        ;;
+    (rpms)
+        cd /gct
+        su builduser -c "/bin/bash -xe /gct/travis-ci/make_source_tarballs.sh"
```
brianhlin: Should we also print the tarball file sizes if we're not running a separate tarballs task?

brianhlin: Should we also print the tarball file size if we're not running a separate tarballs task?

travis-ci/run_task_inside_docker.sh
brianhlin: Do we need `patch` for tasks other than `rpms`?

packaging/fedora/ORDERING
```diff
@@ -0,0 +1,72 @@
...
+globus-gram-client
+globus-gass-server-ez
+globus-gass-cache-program
+globus-xio-gridftp-driver
+globus-gridmap-verify-myproxy-callout
+globus-proxy-utils
+myproxy
+myproxy-oauth
+globus-xioperf
+globus-gatekeeper
```
brianhlin: Do we need to bother building GRAM-related RPMs/tarballs?

travis-ci/run_task_inside_docker.sh
matyasselmeci: Yeah, building and testing gsissh.

packaging/fedora/ORDERING
matyasselmeci: Until a decision is made otherwise, I'm building everything I can.

travis-ci/run_task_inside_docker.sh
matyasselmeci: OK

PR #10, opened 2017-12-19
Deploy to tarballs and SRPMs to GH releases

matyasselmeci:
Build tarballs and SRPMs in Travis-CI and deploy them to GitHub releases.
Building and deploying binary RPMs is also supported but disabled because the
huge number of build products makes manipulating releases unwieldy: uploads
sometimes fail and build products can't be deleted en masse, etc.  I provided a
`Vagrantfile` for someone who wants to build binary RPMs on their own.

The deploy key is for my repo, obviously; I will make a new one for gridcf/gct
once this PR has been merged.

Example: https://github.com/matyasselmeci/gct/releases/tag/v1.1
Build: https://travis-ci.org/matyasselmeci/gct/builds/318843552


travis-ci/make_rpms.sh
```diff
@@ -1,29 +1,34 @@
...
 set -e
+
 usage()
 {
     echo "Usage: $0 [<options>]"
     echo
     echo "Options:"
     echo "  -h                              This message"
     echo "  -C                              Skip %check when building RPMs"
+    echo "  -s                              Stop after building SRPMs"
```
brianhlin: If I was using this script for the first time, this help message would leave me with a few questions: What exactly am I skipping when I stop after building SRPMs? Are there any relevant tasks that happen before building SRPMs? You could either modify this specific sentence or add a description of the tool to the top of the help message to answer these questions.

.travis.yml
```diff
@@ -1,39 +1,66 @@
...
+        - sudo apt-get update
+        - sleep 5
+        - sudo docker pull ${IMAGE}
+
+      script:
+      - travis-ci/setup_tasks.sh
+      deploy:
+        on:
+          tags: true
+          all_branches: true
```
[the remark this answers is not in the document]
matyasselmeci: Nope, that was for testing only. Good catch!

travis-ci/make_rpms.sh
matyasselmeci: I will change it to "Build SRPMs only", which is more obvious.

PR #11, opened 2018-01-18
Also create gct source installer tarball

fscheiner:
This contains all gct source files (w/gsi-openssh patches but w/o .git and
below) with "autoreconf -i" already ran. This is similar to what Globus
shipped as "Source Tarball" from e.g. [1].

[1]: http://toolkit.globus.org/toolkit/downloads/latest-stable/

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,19 +1,9 @@
-m4_define([globus_buildno], [m4_syscmd([
-if test -e .git ; then
-    git log -n 1 --pretty=format:%ct > globus-version.inc.new;
-    if ! cmp globus-version.inc.new globus-version.inc > /dev/null 2>&1; then
-        mv globus-version.inc.new globus-version.inc
-    else
-        rm globus-version.inc.new
-    fi
-fi])dnl
-m4_[]include(globus-version.inc)])dnl
-m4_define([gtreleasebuild], [6.0.]globus_buildno)
-AC_INIT([globus_toolkit], gtreleasebuild, [https://github.com/globus/globus-toolkit/issues])
+m4_define([gtreleasebuild], [6.9.0])
+AC_INIT([gct], gtreleasebuild, [https://github.com/gridcf/gct/issues])
 
 m4_define([short_version], [m4_substr(AC_PACKAGE_VERSION, [0],
     m4_index(AC_PACKAGE_VERSION, [.]))])
-AC_PREFIX_DEFAULT([/usr/local/globus-]short_version)
+AC_PREFIX_DEFAULT([/usr/local/gct-]short_version)
 export GLOBUS_VERSION=$VERSION
 syscmd([./update-dirt.sh])
 syscmd([test -x ./prep-gsissh && sh -x ./prep-gsissh || true])
--- a/travis-ci/make_source_tarballs.sh
+++ b/travis-ci/make_source_tarballs.sh
@@ -36,7 +36,12 @@ rm -rf package-output
 mkdir package-output
 echo '================================================================================'
 sed -i 's/gridftp_hdfs-dist//' Makefile
-make dist && mv gct-*.tar.gz package-output/
+
+# Also create source installer tarball
+package_name=$(grep '^PACKAGE_NAME =' Makefile | cut -d ' ' -f 3)
+package_version=$(grep '^PACKAGE_VERSION =' Makefile | cut -d ' ' -f 3)
+make dist && mv "${package_name}-${package_version}.tar.gz" package-output/
+
 time make -j1 tarballs
 
 echo '================================================================================'
--- a/travis-ci/make_source_tarballs.sh
+++ b/travis-ci/make_source_tarballs.sh
@@ -10,7 +10,6 @@
 
 set -eu
 
-source_installer_tarball="gct-source.tar.gz"
 root=$(git rev-parse --show-toplevel)
 cd "$root"
 
@@ -27,14 +26,6 @@ if [[ ! -f Makefile ]]; then
     if [[ ! -f configure ]]; then
         echo '================================================================================'
         time autoreconf -i
-        if [[ $? -eq 0 ]]; then
-            cd "$root/.."
-            tar -czf $source_installer_tarball --exclude='.git/' "$root"
-            cd "$root"
-        else
-            echo "autoreconf failed. Cannot continue. Exiting" 1>&2
-            exit 1
-        fi
     fi
     echo '================================================================================'
     time ./configure --enable-udt --enable-myproxy --enable-gram5-{server,lsf,sge,slurm,condor,pbs,auditing}
@@ -43,9 +34,9 @@ if [[ ! -f Makefile ]]; then
 fi
 rm -rf package-output
 mkdir package-output
-mv "$root/../$source_installer_tarball" "$root/package-output/"
 echo '================================================================================'
 sed -i 's/gridftp_hdfs-dist//' Makefile
+make dist
 time make -j1 tarballs
 
 echo '================================================================================'
--- a/travis-ci/make_source_tarballs.sh
+++ b/travis-ci/make_source_tarballs.sh
@@ -10,6 +10,7 @@
 
 set -eu
 
+source_installer_tarball="gct-source.tar.gz"
 root=$(git rev-parse --show-toplevel)
 cd "$root"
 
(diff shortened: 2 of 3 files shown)


fscheiner:
In PRACE we still have some HPC centers (including HLRS partly) that installed Globus software like the Globus GridFTP Server from the source tarball and I assume there might be other centers around the world that might prefer this. But there's currently no such source tarball created for gct AFAICS and successfully creating it manually currently requires the existence of the `.git` directory and downloads from external web servers (for gsi-openssh patches) or `autoreconf -i` fails. But downloads from external web servers are often not allowed by default on front end machines, so it would be better if we provide this ready-made.

This should only add the time needed to tar the whole gct dir (w/o `.git` and below) to the build time, as `autoreconf -i` is ran anyhow.

What do you think?

msalle:
Hi,
I think it's a good idea (I use them to for testing on exotic platforms), but would prefer to use something like
 `git archive --format=tar.gz -o gct-source-master.tgz master`
 or so... That would make cleaner tarballs and we could even use specific branches.

fscheiner:
Assuming we will always release from the master branch, what's the benefit of selecting the branch? On the other hand I could think of doing this manually with an argument to `make_source_tarballs.sh` though.

But will `git archive` include the files created by the `autoreconf -i` run done previously? And if we select a specific branch, we also need to checkout that branch prior to `autoreconf -i`.

fscheiner:
Another point: The Travis CI build check failed, but looking into the details ([1], [2], [3]), it failed because some parts failed due to a non-accessible download location during runtime. Can we (as owners or members of the gridcf) retry the Travis CI check manually in such a case?

[1]: https://travis-ci.org/gridcf/gct/jobs/330265112
[2]: https://travis-ci.org/gridcf/gct/jobs/330265113
[3]: https://travis-ci.org/gridcf/gct/jobs/330265114

brianhlin:
@fscheiner if you log into Travis-CI, there should be a button for restarting an entire build or specific tasks in the build

fscheiner:
@brianhlin Thanks, that worked. :D

msalle:
@fscheiner I agree `git archive` is going to miss the `autoreconf -i` output, i.e. it will miss among others `Makefile` and `configure`.
Any reason not to run just `make dist`, it seems to work without problems? I'm not happy with making a tarball manually, that's very bad practise IMHO.

fscheiner:
@msalle Configured Makefiles do not yet exist after running `autoreconf -i`, only `Makefile.in` files. And I believe the source installer tarball should be - and always was - shipped un`configure`d, right?

msalle:
@fscheiner with a "normal" project, you would do something like a `./configure && make dist` or ideally `./configure && make distcheck`. That should give a good clean distribution tarball with `./configure` shipped in it. The user then does the `./configure && make` (without the dist). This does seem to work here too: it results currently in a `globus_toolkit-6.0.1515000527.tar.gz` tarball which looks ok apart from the name: I get for example a `./configure` but no `Makefile`, only `Makefile.am`.

fscheiner:
@msalle OK, I now just use `make dist`, assuming that the resulting tarball will be stored in `dist/`, too. Further changes (maybe to `configure.ac`) are needed for the naming.

fscheiner:
Looks like the naming is defined in the first few lines of [`configure.ac`](https://github.com/gridcf/gct/blob/a75cb7cfb1b61b5e1dbffea4ec69495ddc63db92/configure.ac):
```
m4_define([globus_buildno], [m4_syscmd([
if test -e .git ; then
    git log -n 1 --pretty=format:%ct > globus-version.inc.new;
    if ! cmp globus-version.inc.new globus-version.inc > /dev/null 2>&1; then
        mv globus-version.inc.new globus-version.inc
    else
        rm globus-version.inc.new
    fi
fi])dnl
m4_[]include(globus-version.inc)])dnl
m4_define([gtreleasebuild], [6.0.]globus_buildno)
AC_INIT([globus_toolkit], gtreleasebuild, [https://github.com/globus/globus-toolkit/issues])
[...]
```
Before continuing we maybe should first decide how gct version numbers should look. Or should we continue with what's there already and just change `globus_toolkit` to `gct`?

msalle:
Yes, normally the name set in `AC_INIT` defines the name of the package and tarball. The issues URL in there is shown if you do a `./configure --help`
Concerning the version numbering, I think it makes sense to stick to 6.0.X versions for the time being. Once we're ready to do a production release we can decide whether we jump back to 1.0 (personally I'd prefer not to do that).

fscheiner:
@msalle 
> Concerning the version numbering, I think it makes sense to stick to 6.0.X versions for the time being. Once we're ready to do a production release we can decide whether we jump back to 1.0 (personally I'd prefer not to do that).

I'd also would rather continue with a higher number than with a lower number. I tried to use [Semantic Versioning](https://semver.org/) in the past and I think this would also fit in here:

In general a version number is made of **\<MAJOR\>.\<MINOR\>.\<PATCH\>** there and I'd say our changes do not justify a change to the major version number but to the minor version number. This would also state that gct is compatible with the latest Globus Toolkit in use by users - assuming we didn't or do not introduce any incompatible changes. For patch (level) the time stamp of the last commit could be used as before.

So a gct version number could look like: `6.1.1516867560`

What do others think?

fscheiner:
For this pull request I'd like to continue to develop the change and then put the final version in a new pull request and cancel this one, to not clutter the history with unused and reverted patches.

msalle:
@fscheiner one more thing we probably should take into account: Globus can still bring out new versions, since they still do security patches. How are we going to deal with those concerning versioning? Use the globus version with minor -> minor+1 (assuming we can/want to take over the patch). Concerning the +1: we already have new but backwards compatible features not in the gt...

fscheiner:
@msalle Hard to say. Maybe we wait for what happens and then adapt? As the Grid Community Toolkit is the successor of the Globus Toolkit, I think we should always have a higher version number as the Globus Toolkit for gct. But increasing the version number without code changes but just because the Globus Toolkit uses an equal or higher version than gct is also not justified.

If we keep what Globus uses for version numbers we are "susceptible" to version number collisions, if we change the scheme for version numbers we might still irritate users with similar - between Globus Toolkit and gct - looking version numbers.

I also made an error:
Rereading the [SemVer spec](https://semver.org/#semantic-versioning-specification-semver), using a time stamp as patch level might be conflicting to the spec, because when increasing a minor version, the patch level should be reset to 0.

So a gct version number conforming to the SemVer spec would more look like: `6.1.0`

fscheiner:
Another thought: maybe we start with something close to `7.x.x`, e.g. `6.9.0`, assuming this will always be higher than any Globus Toolkit version numbers until end of 2018 and when we remove something "big" from gct (e.g. GRAM) go to `7.x.x`.

fscheiner:
@msalle Back to the pull request:

When creating the source installer tarball with `make dist`, how can we determine the name of it from [`./travis.ci/make_source_tarballs.sh`](https://github.com/gridcf/gct/blob/master/travis-ci/make_source_tarballs.sh) without parsing `./configure`, `./Makefile` or `./common/source/globus_version`? This because we need to move the resulting tarball (`./gct-?.?.?.tar.gz`) to `./package-output`. I'm currently using a [shell glob pattern](https://github.com/gridcf/gct/pull/11/commits/d50ffad45e00e166a4b3b2fc78d90ab4cba10662#diff-ec7229b4ebd90d294c37378b5acd1708R39) but using the exact name would be better and also not require additional changes if the package name ever changes considerably.

Another way would be to adapt the part regarding the `dist` target in the `./Makefile`. This part is already in `./Makefile.in` (created during `time autoreconf -i`):

```
[...]
dist dist-all: distdir
        tardir=$(distdir) && $(am__tar) | GZIP=$(GZIP_ENV) gzip -c >$(distdir).tar.gz
        $(am__remove_distdir)
[...]
```

...but I don't know how it gets created from `./Makefile.am` or how we can configure in an additional path between `>` and `$(distdir).tar.gz`.

msalle:
@fscheiner first of all, don't change the created targets. If you want to do it by hand, you can use a dist-hook, but I wouldn't do that either. In my experience, sticking to standard autotools as much as possible makes life a lot easier in the long run (and when porting between platforms).
There are a few ways you could do things. If you just need the version number, you can get it from `./configure -V | head -1`. Actually, you can in principle also get our full dist tarball name from there.
But once you have run `configure`, it might be easier to grep for `^PACKAGE_NAME = ` and `^PACKAGE_VERSION = ` in the `Makefile`.

fscheiner:
@msalle OK, I now "parse" the created `./Makefile`.

Version numbers is the remaining issue then. Not sure how to proceed here, maybe a decision of the PMC is needed.

msalle:
Hi,
Another complication with the version numbering is that there are two (or actually many): the individual components also have their own versions. For the gct as a whole, we could consider starting from 7.0.0 since it's not foreseen globus_toolkit wil ever get a 7. Perhaps we should continue this discussion on the mailing list?

fscheiner:
Closed in favour of v2.

PR #30, opened 2018-03-13
Install other devel libs for the xio udt driver so the build won't try to download them

matyasselmeci:
In case we run into the same problem we had with libffi for the other libraries the xio udt driver build might try to download.

### DIFF
--- a/travis-ci/run_task_inside_docker.sh
+++ b/travis-ci/run_task_inside_docker.sh
@@ -34,7 +34,7 @@ packages=(gcc gcc-c++ make autoconf automake libtool \
 
 if [[ $TASK == tests ]]; then
     set +e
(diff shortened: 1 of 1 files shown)


ellert:
This works, but you install some redundant packages.
packages+=(udt-devel libnice-devel)
should be sufficient.
- udt-devel requires udt
- libnice-devel requires glib2-devel
- gettext-devel and libffi-devel are needed to build glib2 from sources, but since glib2 is now installed from packages they are not needed during the GCT build.
- xz was needed to extract the glib2 sources that are xz compressed, but since glib2 is now installed from packages xz is not needed during the GCT build.

matyasselmeci:
Looks like you're right. Ready for re-review.

PR #35, opened 2018-03-31
More gct changes

ellert:
Updates to man pages to say they are GCT, and a few more GT to GCT substitutions.


matyasselmeci:
Ignore the 'check failed' message -- I restarted the broken run and it succeeded the second time.

PR #36, opened 2018-04-03
Bump package versions for GCT release

ellert:


ellert:
1. soname = major version - age. Increasing both major version and age by one means the soname is unchanged.
2. We already bumped the major version (and age) for globus-gridftp-server and globus-gridftp-server-control in GCT due to additions to the API. We do not need to bump the major version again to be greater than the GT major version.

PR #39, opened 2018-04-12
Add BuildRequires: gcc

ellert:
This is to adapt to an upcoming change scheduled for Fedora 29:
https://fedoraproject.org/wiki/Changes/Remove_GCC_from_BuildRoot


ellert:
noarch packages do not need gcc to build (there are a few more).

Issue #47, opened 2018-08-07
Bug in /etc/init.d/globus-gridftp-server affects Pacemaker


adorsey-NOAA:
I've found a bug in the init script for the GridFTP server in current versions of the globus-gridftp-server-progs RPM for RHEL 6.  Not sure if that's been fixed in your fork, but I wanted to make you aware in case.

Whenever this script is run with the status option (/etc/init.d/globus-gridftp server status) the status is correctly reported, but the return code is always 0.  This isn't a problem until someone tries to use Pacemaker to manage GridFTP services.  The incorrect return code causes Pacemaker to always think the GridFTP service is running, so it never gets started.

The following hack makes the script return an exit value correctly when the status parameter is passed.

```
--- globus-gridftp-server.orig  2018-08-07 22:56:34.217826950 +0000
+++ globus-gridftp-server       2018-08-07 22:58:17.035903704 +0000
@@ -139,6 +139,7 @@
        ;;
     status)
        status
+       rc=$?
        ;;
     restart | force-reload)
        restart
```

ellert:
This was addressed by merging the corresponding fix from the old GT upstream.

PR #48, opened 2018-08-08
Fix exit code of status command in the gridftp service

brianhlin:
Addresses #47

### DIFF
--- a/gridftp/server/src/init/globus-gridftp-server.in
+++ b/gridftp/server/src/init/globus-gridftp-server.in
@@ -183,19 +183,20 @@ status()
         if [ "$pid" != "" -a "$pid" -gt 0 ]; then
             if kill -0 "$pid" 2> /dev/null; then
                 echo "GridFTP server is running (pid=$pid)"
-                return 0
+                rc=0
             else
                 echo "Stale PID file $pidfile"
-                return 1
+                rc=1
             fi
         else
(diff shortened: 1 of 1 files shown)


matyasselmeci:
This code works because $rc in that function is actually global so it gets propagated to the caller, but you need to read it closely to realize that. I suggest changing the case statement like in #47.

fscheiner:
Just had a quick look: I think the problem described in #47 is due to missing LSB functions at the OP's machine, so instead of `status_lsb` only `status` is called which does not set `$rc` but just returns a value, but the `exit $rc` at the end of the "main" `case` clause doesn't anticipate this.

As all functions return a value (either the value of "$rc" or just a value), the easiest solution might be to change the last `exit` call to `exit $?` (which will exit with the return or exit value of the last called function or tool) or just `exit` (which will act the same way).

What do you think about that?

brianhlin:
Personally, I think it's nice to explicitly state `exit $exit_code` but I'm also fine with `exit $?` at the end, too.

brianhlin:
Note that the commit (https://github.com/gridcf/gct/commit/cef7e1e3a2bdc93631cf2c25a947cfd07fc76b1a) fixes it depending on the global `rc` variable, which we decided against in this PR. However, I think for such a minor fix it doesn't warrant diverging away from upstream

PR #49, opened 2018-08-28
Deploy released tarballs to repo.gridcf.org

matyasselmeci:
The Travis script will upload tagged releases to repo.gridcf.org via SFTP. A successful run can be seen [here](https://travis-ci.org/matyasselmeci/gct/builds/421789275) with the results at http://repo.gridcf.org/gct6/sources/.


matyasselmeci:
I'll experiment with a few things to see if we can get that behavior.

matyasselmeci:
OK, took a while but I implemented something. The script checks if remote tarballs of the same name already exist, then downloads them and runs some verifications. If the remote tarball is OK, then it doesn't upload the newly built tarballs.

I had to store the checksums as separate `.sha512` files instead of one big `sha512sums` file because the build script only knows about the new build products, not any older ones that are in the repos. I hope that's OK.

fscheiner:
Sorry to re-open this PR, it slipped though and I have some questions and/or like to discuss a few things. I'll comment on the code if applicable.

fscheiner:
Well I actually wanted to see the commented code portion in this discussion. Maybe I should have commented on the code in @matyasselmeci 's fork then. But if the `deploytorepo` branch is removed there later, won't this comment vanish, too? I don't know, but I think you'll still get my point.

fscheiner:
I'm aware that this slightly contradicts my argument in https://github.com/gridcf/gct/pull/49#commitcomment-30578289, but IMHO it also makes sense to update the full source tarball in such a case. So maybe one full source tarball per release plus one latest full source tarball after any change that always overwrites the previous latest full source tarball?

PR #50, opened 2018-09-04
Split guc-cc.pl test into smaller tests

ellert:
The guc-cc.pl in the globus-gass-copy package take a long time to run on slow computers, and after the increase of the RSA key size used for testing it takes even longer.

The official Debian build servers have a test to detect stuck builds that kills a build if it hasn't printed anything to stdout/stderr for 150 minutes (2½ hours). When running the test suite the output from the test is redirected to various log files and only at the end of the test a status is written to stdout.

With the longer execution time due to the larger key size, the execution time of the guc-cc.pl test exceeds 2½ hours on the slower Debian build servers (has happened on hppa, armhf and mips builds).

This PR splits the guc-cc.pl into 6 smaller tests. The same tests are performed as with the original version, but since it is split into smaller tests the time between the output of the test status line is cut by a factor 6 and the no-activity test is not triggered on the slower build servers.

### DIFF
--- a/gass/copy/source/configure.ac
+++ b/gass/copy/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gass_copy],[10.2],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_gass_copy],[10.3],[https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
 AC_SUBST([MAJOR_VERSION], [${PACKAGE_VERSION%%.*}])
 AC_SUBST([MINOR_VERSION], [${PACKAGE_VERSION##*.}])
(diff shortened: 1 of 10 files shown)


msalle:
even split up, it's still a massive amount of time. Any thoughts on whether this test is actually useful?
Not sure whether doing concurrency tests and the like on virtualized and probably not too fast hardware is very accurate for such a test.

just my 2ct...

ellert:
The build times (including testing) for globus-gass-copy:
https://buildd.debian.org/status/logs.php?pkg=globus-gass-copy
are of the same order as those for globus-ftp-client:
https://buildd.debian.org/status/logs.php?pkg=globus-ftp-client
Just globus-ftp-client didn't cram all the tests into a single test script.
For most of the architectures the build time is not very long, it is really only hppa that is extreme.
Some of the architectures by default builds with the "nocheck" option. I don't think we need to do our own selection of what tests to run in addition to that.

msalle:
That's clear. In any case, I've approved the PR.
Concerning disabling the slowest tests on certain architectures, that makes a lot of sense, but probably  requires looking at the type of test in detail.

PR #55, opened 2018-09-21
Allow TLS 1.0/1.1

ellert:
The old GT upstream changed the default minimum TLS version form 1.0 to 1.2 in their globus-gssapi-gsi version 13.9 (merged into GCT as our version 14.6).

There have been reports that this causes compatibility issues with some servers used in production in WLCG.

* https://ggus.eu/index.php?mode=ticket_info&ticket_id=137130
* https://mailman.egi.eu/pipermail/gt-eos/2018-September/000178.html

It was suggested that this change should be reverted, at least temporary. This PR sets the minimum TLS version to 1.0 in the gsi.conf configuration file.

### DIFF
--- a/gsi/gssapi/source/configure.ac
+++ b/gsi/gssapi/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gssapi_gsi],[14.7],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_gssapi_gsi],[14.8],[https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
 AC_SUBST([MAJOR_VERSION], [${PACKAGE_VERSION%%.*}])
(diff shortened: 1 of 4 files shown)


maarten-litmaath:
Hi Mattias,
re-enabling TLS 1.0 and 1.1 for the time being would allow WLCG to look into the
adaptation of legacy services that for some reason do not support TLS 1.2 (yet).
Thanks very much for your efforts in these matters!

ellert:
On th GGUS ticket the admin on the site says they are using globus-gssapi-gsi 11.22-1. This version is quite old (September 2015).

In versions before 11.26 (January 2016) there was a bug that meant that using the FORCE_TLS option meant forcing TLS 1.0, i.e. it also disabled TLS 1.1 and 1.2, and no only SSLv3.

Starting from version 12.15 (April 2017) support for SSLv3 was dropped, and only TLS allowed.

How common is it for sites to have globus-gssapi-gsi versions before 11.26 installed and also setting the FORCE_TLS option in gsi.conf?

PR #57, opened 2018-11-05
Bump GCT release version to 6.2

ellert:


packaging/debian/globus-scheduler-event-generator/debian/rules
```diff
@@ -27,8 +27,6 @@ configure-stamp:
 
 	dh_autoreconf
 
-	GLOBUS_VERSION=6.0 \
-	GRIDMAP=/etc/grid-security/grid-mapfile \
 	dh_auto_configure -- \
```
matyasselmeci: How come you're not setting `GLOBUS_VERSION=6.2` here too?

ellert: The configure script for globus-scheduler-event-generator does not mention GLOBUS_VERSION, so it does not need to be set. The globus-scheduler-event-generator.spec file for fedora/rhel does not set it. Removing it in debian/rules makes it consistent.

fscheiner:
@matyasselmeci @ellert @msalle 
Now that Mattias' changes are included, shouldn't we create a tag (e.g. `v6.2`, `v6.2.0`, `6.2`, `6.2.0`), so new source packages are produced?

msalle:
Sounds like a good idea.

Probably good to make tags starting with a `v` and branches without. So to make a tag `v6.2`.

I think we can't really use the 'patch' level since it's now used with a timestamp.

matyasselmeci:
Can we try and fix the 'patch' level issue? I'm not sure how we will be able to tag later 6.2 releases without that.

ellert:
Do we really need a branch for the 6.2 release? The old upstream did not have a special branch for the 6.0 release - or rather they had a globus-6-branch, but no trunk, and the globus-6-branch was the default branch which was slightly confusing - in practise it was just a trunk with a non-standard name. Is anyone planning to merge anything to trunk that will not be released as an update to 6.2? To have a 6.2 branch that always will be the same as trunk is just extra work.

The old upstream stopped using the patch level in the toolkit version number with release 6.0. It was always called just 6.0 after that, even after updates were released. That the toolkit version doesn't change every time something changes was a great simplification, and something we should keep. This of course means that the toolkit version number (6.2) is not suitable to use as the tag name in git on its own.

I think a tag like v6.2.20181110 would make sense, and it also fits nicely with the existing prerelease tags. Using 6.2.0, 6.2.1, etc would be confusing since they look like version numbers, and the toolkit version is 6.2, so what is it that is version 6.2.0?

fscheiner:
@matyasselmeci @ellert @msalle 
I'd like to create a tag for today following Mattias' proposal (i.e. `v6.2.20181115`) so `repo.gridcf.org` receives the latest source packages versions. This because currently the RPM spec files already relate to newer versions of the source packages that are available in the repo and I don't want to change too much in the spec files for usage on the [OBS](https://build.opensuse.org/) to be able to build with the old source packages currently available.

@matyasselmeci 
If need be, I assume we can delete the tag afterwards and the packages will stay, right?

****

Do you all agree?

matyasselmeci:
No objections from me. And yes, deleting the tag will not remove the tarballs from repo.gridcf.org, though I can't think of a good reason to delete the tag in the first place; we haven't gotten rid of the prerelease tags either.

msalle:
It's fine with me, so just go ahead with making a tag.

fscheiner:
@msalle 
Great.

@ellert 
Sorry, couldn't wait for your answer, but assume you would also agree. :-)

****

Publishing the `v6.2.20181115` tag now.

fscheiner:
@matyasselmeci @msalle @ellert 
Just noticed, that the tests on Travis-CI (https://travis-ci.org/gridcf/gct/builds/455701736) are also running.

Is this really needed when we create a tag? Can't we consider the code base always tested when we create a tag? This because no addition is or should be made if the tests on Travis-CI fail for a non-service related error, so we never should have a situation where a tag points to an untested code base.

Hm, rethinking this, maybe the above consideration is only true if the created tag points to HEAD or merge commits. But OTOH I currently can't think of any reason why to tag something not HEAD or a merge commit.

matyasselmeci:
@fscheiner, Travis-CI is also used to run the scripts that deploy the tarballs to GitHub and repo.gridcf.org.

fscheiner:
@matyasselmeci 
Maybe I misunderstood things, long time since I had a look into the Travis-CI configurations. So the source tarballs are created during the Test stage and "only" deployed during the Deploy stage? Makes sense. :-)

fscheiner:
@matyasselmeci 
Hm, it currently keeps restarting the Deploy stage after erroring out. First time I saw a "bad gateway" error. Is this a temporary error?

matyasselmeci:
It's managed to upload some of the files; the last one it failed at was the myproxy-oauth RPM. I was the one that restarted it; I want to see if it gets any further.

matyasselmeci:
It just seems to fail at a random file.
@bbockelm, since we have repo.gridcf.org, it seems superfluous to also upload all the individual tarballs and SRPMs to the GitHub release page. Perhaps the big gct tarball would be good enough. What do you think?

matyasselmeci:
OK, after a couple of retries, it finally succeeded and we have a full set of files on GitHub releases and in https://repo.gridcf.org/gct6/sources/.

msalle:
It seems to publish only the tar balls to repo.gridcf.org, not the SRPMs. Is that intentional?

ellert:
Updated packages available in Fedora updates testing, Fedora rawhide, EPEL testing and Debian unstable.

PR #61, opened 2018-11-28
Fix FTBR (failed to build reproducibly)

ellert:
Of the GCT packages globus-common is the only one that gets flagged for FTBR issues:
https://tests.reproducible-builds.org/debian/globus-common
This PR addresses these issues.


packaging/debian/globus-common/debian/rules
```diff
@@ -30,6 +30,7 @@ configure-stamp:
 	dh_autoreconf
 
 	GLOBUS_VERSION=6.2 \
+	SH=/bin/sh \
```
matyasselmeci: Do you also need to define `PERL`?

ellert: Not according to the [logs in the report](https://tests.reproducible-builds.org/debian/dbdtxt/unstable/amd64/globus-common_18.0-2.diffoscope.txt.gz):
```
│ │ │ │  # Interpreters:
│ │ │ │  GLOBUS_SH_PERL="/usr/bin/perl"
│ │ │ │ -GLOBUS_SH="/bin/sh"
│ │ │ │ +GLOBUS_SH="/usr/bin/sh"

```

ellert: The man page for globus-sh-exec says that GLOBUS_SH_PERL and GLOBUS_SH will de defined when scripts are run using that wrapper, so not defining them there would break things relying on that documented behaviour.

msalle: Ah, right, missed that, actually had missed how they were used. After a closer look at the script and manpage your fix seems perfectly fine.

Issue #25, opened 2018-03-01
Remove usage statistics reporting


matyasselmeci:
According to http://toolkit.globus.org/toolkit/docs/latest-stable/Usage_Stats.html, GT phones home with usage statistics. We should remove that.

After removing the code, also update the documentation to state that GCT no longer has usage statistics reporting.

ellert:
See proposed changes in PR #63.

PR #63, opened 2018-12-07
Remove usage statistics collection support

ellert:
This PR drops the usage statistics collection feature from globus-gram-job-manager, globus-gridftp-server and myproxy. With this change they will no longer depend on the globus-usage package. See issue #25.


matyasselmeci:
This looks good from a packaging perspective but I'd like a C programmer to review the code changes before merging it. @msalle , @fscheiner , could one of you take a look?

PR #69, opened 2019-02-18
Add su option to logrotate file

ellert:
This addresses failures to start logrotate on newer releases that requires the su option when the directory containing the files to be rotated have less strict access permissions:
```
● logrotate.service - Rotate log files
   Loaded: loaded (/lib/systemd/system/logrotate.service; static; vendor preset: enabled)
   Active: failed (Result: exit-code) since Wed 2019-02-13 15:40:11 CET; 46min ago
     Docs: man:logrotate(8)
           man:logrotate.conf(5)
  Process: 290 ExecStart=/usr/sbin/logrotate /etc/logrotate.conf (code=exited, status=1/FAILURE)
 Main PID: 290 (code=exited, status=1/FAILURE)

feb 13 15:40:02 debian-unstable systemd[1]: Starting Rotate log files...
feb 13 15:40:09 debian-unstable logrotate[290]: error: skipping "/var/log/globus/gram_*.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
feb 13 15:40:11 debian-unstable systemd[1]: logrotate.service: Main process exited, code=exited, status=1/FAILURE
feb 13 15:40:11 debian-unstable systemd[1]: logrotate.service: Failed with result 'exit-code'.
feb 13 15:40:11 debian-unstable systemd[1]: Failed to start Rotate log files.
```

### DIFF
--- a/gram/jobmanager/source/configure.ac
+++ b/gram/jobmanager/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gram_job_manager],[15.3],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_gram_job_manager],[15.4],[https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
 AC_SUBST([MAJOR_VERSION], [${PACKAGE_VERSION%%.*}])
 AC_SUBST([MINOR_VERSION], [${PACKAGE_VERSION##*.}])
--- a/gram/jobmanager/source/gram.logrotate.in
+++ b/gram/jobmanager/source/gram.logrotate.in
@@ -8,4 +8,5 @@
     rotate 4
     sharedscripts
     copytruncate
+    su root root
 }
--- a/packaging/debian/globus-gram-job-manager/debian/changelog.in
+++ b/packaging/debian/globus-gram-job-manager/debian/changelog.in
@@ -1,3 +1,9 @@
+globus-gram-job-manager (15.4-1+gct.@distro@) @distro@; urgency=medium
+
+  * Add su option to logrotate file
+
+ -- Mattias Ellert <mattias.ellert@physica.uu.se>  Fri, 15 Feb 2019 16:11:09 +0100
+
 globus-gram-job-manager (15.3-1+gct.@distro@) @distro@; urgency=medium
 
   * Remove usage statistics collection support
--- a/packaging/fedora/globus-gram-job-manager.spec
+++ b/packaging/fedora/globus-gram-job-manager.spec
@@ -2,7 +2,7 @@
 
 Name:		globus-gram-job-manager
 %global _name %(tr - _ <<< %{name})
-Version:	15.3
+Version:	15.4
 Release:	1%{?dist}
 Summary:	Grid Community Toolkit - GRAM Jobmanager
 
@@ -115,6 +115,11 @@ make install DESTDIR=$RPM_BUILD_ROOT
 # Remove libtool archives (.la files)
 rm $RPM_BUILD_ROOT%{_libdir}/*.la
 
+%if %{?rhel}%{!?rhel:0} == 6
+# Remove su option from logrotate file in EPEL 6 (not supported)
+sed '/ su /d' -i $RPM_BUILD_ROOT%{_sysconfdir}/logrotate.d/globus-job-manager
+%endif
+
 %check
 GLOBUS_HOSTNAME=localhost make %{?_smp_mflags} check VERBOSE=1
 
@@ -153,6 +158,9 @@ GLOBUS_HOSTNAME=localhost make %{?_smp_mflags} check VERBOSE=1
 %{_libdir}/libglobus_seg_job_manager.so
 
 %changelog
+* Fri Feb 15 2019 Mattias Ellert <mattias.ellert@physics.uu.se> - 15.4-1
+- Add su option to logrotate file
+
 * Fri Dec 07 2018 Mattias Ellert <mattias.ellert@physics.uu.se> - 15.3-1
 - Remove usage statistics collection support
 


ellert:
What user would you use instead? As far as I know the services themselves run as root and change user when they run subprocesses on behalf of different users depending ot the user mapping in the grid map file.
The default user doing the rotation in logrotate is root. With olders version of logrotate the rotations in this config file were run by root. Newer versions refuse to use the default user in some circumstances. This change tells these new version to do what the older version already are doing.
There are no other lines in the file matches ' su ', and I think that ' su ' is easier to read than '^[[:space:]]+su[[:space:]]'. For this sed command to start doing anything it is not supposed to do someone must check in changes to the file in git that adds additional lines that match ' su '.

msalle:
> What user would you use instead? As far as I know the services themselves run as root and change user when they run subprocesses on behalf of different users depending ot the user mapping in the grid map file.
> The default user doing the rotation in logrotate is root.

If the rotation is done by root, it indeed has to be root.
We noticed (from local sysadmins) that some of the gridftp-servers, e.g. those run by CREAM and DPM, are running under special accounts, so not just after the user mapping has taken place, but the main process. If this is not a generic pattern, or if there is no clear default, then root is again indeed the only option.

> There are no other lines in the file matches ' su ', and I think that ' su ' is easier to read than '^[[:space:]]+su[[:space:]]'. For this sed command to start doing anything it is not supposed to do someone must check in changes to the file in git that adds additional lines that match ' su '.

It's true that it's easier to read and given that this is our upstream logrotate file I agree that we can say it's the only line. It's just that with the matching from the start of the line we have less chance of accidentally removing a comment if one gets ever inserted.
In either case, I don't want to block the pull request, but I thought it was good to clarify these points first.

fscheiner:
@ellert @msalle 
When adapting the RPM spec files for the SLES/OpenSUSE package builds I also stumbled across this issue, though not knowing it also affects Debian (unstable). And it actually was brought to my attention by the build system which also complained about directories being world-writable.

This should be fixed, because I assume as long as the logfile-creating service hasn't started, ordinary users could precreate the matching logfiles and gain access to the contents this way when the service starts, except if the logfile-creating service checks for this scenario. Even then we would allow ordinary users to create files in a log directory.

As I didn't want to changed the source packages I used the following to correct the permissions (also for another dir):
```
[...]
%install
make install DESTDIR=$RPM_BUILD_ROOT

# Remove libtool archives (.la files)
rm $RPM_BUILD_ROOT%{_libdir}/*.la

+# Fix permissions (dirs are actually installed world-writable!)
+chmod 0755 $RPM_BUILD_ROOT%{_localstatedir}/log/globus
+chmod 0755 $RPM_BUILD_ROOT%{_localstatedir}/lib/globus/gram_job_state

%check
GLOBUS_HOSTNAME=localhost make %{?_smp_mflags} check VERBOSE=1
[...]
```

I have to double-check, but I seem to remember this was also needed for some other package(s). I'll update you on this topic if I find another affected package.

We should maybe better change that in the build scripts, though I don't yet know where exactly.

fscheiner:
@ellert @msalle 
Ok, looks like this is limited to GRAM and the second package is `globus-gram-audit`:
```
~/osc/home:frank_scheiner:gct> grep -nri --exclude-dir=.osc -A 3 '# Fix permissions' *
globus-gram-audit/globus-gram-audit.spec:58:# Fix permissions (dir is actually installed world-writable!)
globus-gram-audit/globus-gram-audit.spec-59-chmod 0755 $RPM_BUILD_ROOT%{_localstatedir}/lib/globus/gram-audit
globus-gram-audit/globus-gram-audit.spec-60-
globus-gram-audit/globus-gram-audit.spec-61-%post
--
globus-gram-job-manager/globus-gram-job-manager.spec:130:# Fix permissions (dirs are actually installed world-writable!)
globus-gram-job-manager/globus-gram-job-manager.spec-131-chmod 0755 $RPM_BUILD_ROOT%{_localstatedir}/log/globus
globus-gram-job-manager/globus-gram-job-manager.spec-132-chmod 0755 $RPM_BUILD_ROOT%{_localstatedir}/lib/globus/gram_job_state
globus-gram-job-manager/globus-gram-job-manager.spec-133-
```

msalle:
I might be missing something, but isn't the normal way to add the directory explicitly in the list of files and then prefix it with an ``%attr`` so something like
`` %attr(755, gram, root) %{_localstatedir}/log/globus``
I think the same also applies to files themselves actually.

ellert:
The access permissions on the directories in the packages are the way they are on purpose.
It is not a packaging mistake.
The access permissions reflect the permissions as set on these directories when they are created by "make install".
Since the author of the makefile made an effort to set a non-standard permission on these directories, there must be a reason for it. So changing it to the default permissions will most likely make some things no longer work.
Package checking tools like rpmlint and lintian are good at detecting potential packaging mistakes, and they are very helpful. But if you follow them indiscriminately and make all their complaints go away you risk introducing changes that make the packages broken.

fscheiner:
@ellert 
> Package checking tools like rpmlint and lintian are good at detecting potential packaging mistakes, and they are very helpful. But if you follow them indiscriminately and make all their complaints go away you risk introducing changes that make the packages broken.

Though I tend to agree with this (as I don't have that much experience with RPM packaging, to be honest), I also don't see a valid reason for a setuid() using service to need a world-writable directory (even with sticky bit set).

I did some "research" and as per the [GRAM5 admin documentation about Security Considerations](https://gridcf.org/gct-docs/gram5/admin/index.html#gram5-admin-security_considerations), the `globus-gatekeeper` runs as root and uses setuid() to first change to the authenticated (unprivileged) user and then start an instance of `globus-job-manager`. To allow this instance r/w access to a root owned directory (`/var/log/globus` or perhaps also multiple directories, as directories below `/var/lib/globus` are also world-writable (see https://github.com/gridcf/gct/pull/69#issuecomment-465139497), that directory "needs" to be world-writable.

An alternative solution would be for the `gatekeeper` to pre-create a subdirectory with the needed permissions and ownership for the authenticated user where the `globus-job-manager` instance can operate in, before doing the setuid(). That subdirectory cannot be removed by the `globus-job-manager` instance, but it can operate freely inside this subdirectory, i.e. create and destroy necessary files "at will". All that without the need for a world-writable directory.

If the effort to make such changes is justified if GRAM5 is already no longer used widely is the question, though. Maybe we just remove GRAM5 sometime in the future after asking the community about their needs with it and be happy without it later on.

@msalle 
To make these packages work as intended in the meantime, is it possible to overrule the rpmlint _badness_ results? IIRC the OBS build system refused to build the packages without changing the FS permissions for the mentioned dirs.

msalle:
In general: let's not spend much time on GRAM5, I think NO one is using it any longer...
Secondly, world-writeable directories are not a good idea, so if we can somehow prevent that, it would be very good. If logrotate does something as root that's less problematic.
Also, why doesn't it log to syslog? Users shouldn't be logging to logfiles, only special services. Syslog is the thing for that, since you can write, but not change.

@fscheiner yes you can, you need to add a file PACKAGENAME-rpmlintrc which then contains something like
`addFilter("`REGEXP MATCHING THE ERROR MESSAGE`")`
See e.g. https://build.opensuse.org/package/view_file/home:msalle:sac/lcmaps/lcmaps-rpmlintrc
I think you have to add it also as source to the spec file.

fscheiner:
@msalle:
> yes you can, you need to add a file PACKAGENAME-rpmlintrc which then contains something like
addFilter("REGEXP MATCHING THE ERROR MESSAGE")
See e.g. https://build.opensuse.org/package/view_file/home:msalle:sac/lcmaps/lcmaps-rpmlintrc
I think you have to add it also as source to the spec file.

Thanks, that worked like a charm, see for example https://build.opensuse.org/package/show/home:frank_scheiner:gct/globus-gram-job-manager

msalle:
Just two questions: was it a fatal error? If just a warning, I would not suppress it but leave it in. Secondly,isn't it just the `gram_job_state` file that needs to be world-writeable? Is it possible to pre-create that, and leave the directory itself non-world-writeable? That could probably also be fixed to survive a logrotate (which still needs to run as root).

fscheiner:
@msalle:
> [...] was it a fatal error? If just a warning, I would not suppress it but leave it in.

Unfortunately world-writable permissions are evaluated as error with a _badness_ of 10000. As there are two world-writable dirs in the mentioned package, the result is a total _badness_ of 20000 and IIRC a maximum of 1000 is allowed.

msalle:
Ah, well then there is little to do about that. Any thoughts on the pre-creating of the logfile? You can have logrotate create a new one with permissions (and ownership), see create in man logrotate.conf

fscheiner:
See [my comment from 2019-02-21](https://github.com/gridcf/gct/pull/69#issuecomment-466026132) for reference: I think the names of the files that will be created in these world-writable directories are not known during package installation, but only during runtime of the service:

From https://gridcf.org/gct-docs/gram5/admin/index.html#gram5-admin-security_considerations:
> #### Job Manager Security Considerations ####
> [...]
> The globus-job-manager program uses a publicly-writable directory for job state files. This directory has the sticky bit set, so users may not remove other users files. Each file is named by a UUID, so it should be unique.

From that I'd say the documentation refers to the `/var/lib/globus/gram_job_state` directory. But as the `/var/log/globus` directory is also world-writable (see https://github.com/gridcf/gct/pull/69#issuecomment-465139497), I assume that `globus-job-manager` also places files there, possibly in a similar manner.

In addition the [`gram.logrotate` file](https://github.com/gridcf/gct/blob/master/gram/jobmanager/source/gram.logrotate.in) uses `@LOGDIR@/gram_*.log`, which I interpolate to that the log files could be named `gram_<UUID>.log`, so can't be pre-created during package installation.

Hence I think it can only be "fixed" in the service itself.

msalle:
Hi,

On Wed, 27 Feb 2019, 17:18 fscheiner <notifications@github.com wrote:

> See my comment from 2019-02-21
> <https://github.com/gridcf/gct/pull/69#issuecomment-466026132> for
> reference: I think the names of the files that will be created in these
> world-writable directories are not known during package installation, but
> only during runtime of the service:
>
> From
> https://gridcf.org/gct-docs/gram5/admin/index.html#gram5-admin-security_considerations
> :
>
> Job Manager Security Considerations
>
> [...]
> The globus-job-manager program uses a publicly-writable directory for job
> state files. This directory has the sticky bit set, so users may not remove
> other users files. Each file is named by a UUID, so it should be unique.
>
> From that I'd say the documentation refers to the
> /var/lib/globus/gram_job_state directory. But as the /var/log/globus
> directory is also world-writable (see #69 (comment)
> <https://github.com/gridcf/gct/pull/69#issuecomment-465139497>), I assume
> that globus-job-manager also places files there, possibly in a similar
> manner.
>
> In addition the gram.logrotate file
> <https://github.com/gridcf/gct/blob/master/gram/jobmanager/source/gram.logrotate.in>
> uses @LOGDIR@/gram_*.log, which I interpolate to that the log files could
> be named gram_<UUID>.log, so can't be pre-created during package
> installation.
>
> Hence I think it can only be "fixed" in the service itself.
>

Indeed, you're right, see the option -log-pattern in the man page...
Ok, so there isn't anything we can do about it...

PR #70, opened 2019-02-27
Make %_name definition be POSIX-compatible

matyasselmeci:
The `tr` is run with /bin/sh so on systems where /bin/sh is not bash, rpmbuild -bp fails.

### DIFF
--- a/packaging/fedora/globus-authz-callout-error.spec
+++ b/packaging/fedora/globus-authz-callout-error.spec
@@ -2,7 +2,7 @@
 
 Name:		globus-authz-callout-error
 %global soname 0
-%global _name %(tr - _ <<< %{name})
+%global _name %(echo %{name} | tr - _)
 Version:	4.1
 Release:	1%{?dist}
 Summary:	Grid Community Toolkit - Globus authz error library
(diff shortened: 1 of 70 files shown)


fscheiner:
Good catch! Always thought this was some sort of RPM macro magic. Out of interest, on what OS did you stumble upon this issue?

matyasselmeci:
My coworker was using Ubuntu, where /bin/sh is a symlink to /bin/dash.

fscheiner:
Ok, so did your coworker try to build the RPMs on Ubuntu?

matyasselmeci:
No, he just wanted to extract the spec files from the SRPMs by running `rpmbuild -bp` (which stops after the `%prep` step).

fscheiner:
Good test for POSIX compatibility. :-D

Looks good to me, btw.

msalle:
assuming the `tr <<<` into `echo ...| tr` is the only change, I fully approve too. There was a similar discussion for OpenSUSE recently, whether it could be assumed that `/bin/sh` is `bash` (most people incl. me agreed when using `/bin/sh` you should not use bashisms or other non-POSIX features).
Like Ubuntu Debian also has `/bin/dash` (reverse actually, Ubuntu follows Debian).

PR #73, opened 2019-04-02
Eppn force map

stdweird:
This is the same PR as https://github.com/globus/globus-toolkit/pull/126
I was advised by @msalle to make the PR also here, but i hvae no idea how this will conflict with any later backports/rebases
@msalle also mentioned to add some documentation, but i'm a bit clueless where i should do so

### DIFF
--- a/gsi/gridmap_eppn_callout/globus_gridmap_eppn.c
+++ b/gsi/gridmap_eppn_callout/globus_gridmap_eppn.c
@@ -1,12 +1,12 @@
 /*
  * Copyright 1999-2006 University of Chicago
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -32,7 +32,7 @@
 #endif
 
 #if OPENSSL_VERSION_NUMBER < 0x0090801fL
-#define GT_ASN1_GET_OBJECT_CAST 
+#define GT_ASN1_GET_OBJECT_CAST
 #else
 #define GT_ASN1_GET_OBJECT_CAST (const unsigned char **)
 #endif
@@ -40,7 +40,7 @@
 
 /* 1.2.3.4.4.3.2.1.7.8 */
 static const gss_OID_desc               ggvm_cert_chain_oid =
-    {11, "\x2b\x06\x01\x04\x01\x9b\x50\x01\x01\x01\x08"}; 
+    {11, "\x2b\x06\x01\x04\x01\x9b\x50\x01\x01\x01\x08"};
 
 /* 1.3.6.1.4.1.5923.1.1.1.6 */
 static const char*                      ggvm_eppn_oid =
@@ -119,7 +119,7 @@ ggvm_extract_cert_from_chain(
     while(!cert && cert_index < cert_chain_buffers->count)
     {
         globus_gsi_cert_utils_cert_type_t   cert_type;
-        
+
         ptr = cert_chain_buffers->elements[cert_index].value;
         cert = d2i_X509(
             NULL,
@@ -174,7 +174,7 @@ err:
 
 
 static
-globus_result_t 
+globus_result_t
 ggvm_verify_cert(
     X509 *                              cert,
     X509 *                              ca_cert,
@@ -239,7 +239,7 @@ ggvm_verify_cert(
             cert_ctx, X509_V_FLAG_USE_CHECK_TIME);
         X509_STORE_CTX_set_time(cert_ctx, 0, shared_exp - 1);
     }
-    
+
     rc = X509_verify_cert(cert_ctx);
     if(rc != 1)
     {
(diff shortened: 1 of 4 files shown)


fscheiner:
@stdweird 
Regarding  the documentation: I would have pointed you to the corresponding manpage but it looks like there is none or it is documented somewhere else. In addition we also maintain a separate documentation mainly intended for online reading at https://gridcf.org/gct-docs/ (corresponding git repo at https://github.com/gridcf/gct-docs/) but I also don't see any mention of "eppn" there when using the GitHub search function.

So maybe for a start you could explain (best in your first comment) what your changes do accomplish and why these are useful/needed for the GCT.

In addition a description of how this functionality works and can be used would be useful and could be added to the GCT documentation later on. Maybe you can have a read in the GCT documentation on some similar topics and come up with some documentation in the future?

@msalle:
Does this meet your requirements? As you already reviewed these changes for the Globus Toolkit, could you also review and possibly approve these for the GCT?

gsi/gridmap_eppn_callout/globus_gridmap_eppn.c
```diff
@@ -615,24 +616,45 @@ globus_gridmap_eppn_callout(
...
-                result,
-                GLOBUS_GRIDMAP_CALLOUT_LOOKUP_FAILED,
-                ("Credentials specify id of %s, can not allow id of %s.\n",
-                 found_identity, desired_identity));
-            globus_free(found_identity);
-            goto error;
+            if(getenv("GLOBUS_EPPN_FORCE_GRIDMAP"))
+            {
+                result = globus_gss_assist_gridmap(found_identity, &tmp_identity);
+                if(result != 0)
```
ellert: globus_gss_assist_gridmap returns int. "result" is a globus_result_t, so this does not match. Should be
rc = ...
if (rc != 0)

gsi/gridmap_eppn_callout/globus_gridmap_eppn.c
```diff
@@ -615,24 +616,45 @@ globus_gridmap_eppn_callout(
...
     }
     else
     {
         /* proceed with gridmap lookup */
         if(desired_identity == NULL)
         {
-            rc = globus_gss_assist_gridmap(subject, &found_identity);
-            if(rc != 0)
+            result = globus_gss_assist_gridmap(subject, &found_identity);
+            if(result != 0)
```
ellert: globus_gss_assist_gridmap returns int, so "rc" as before this change is correct. "result" is globus_result_t and doesn't match.

gsi/gridmap_eppn_callout/globus_gridmap_eppn.c
msalle: Ah yes, you're totally right. So it should be a different variable indeed. But then the question is what to pass to `GLOBUS_GRIDMAP_CALLOUT_ERROR()`:
Going through the whole file, I notice that `GLOBUS_GRIDMAP_CALLOUT_ERROR()` is often called with `result == GLOBUS_SUCCESS`, e.g. in `ggvm_load_cert_from_file()`, while in other cases it's actually initialized to `GLOBUS_FAILURE`, as in `ggvm_verify_cert()` ?! I'd say `GLOBUS_FAILURE` is the logical one, but not in line with what most of the rest of the file does.
The second case below is easier, there we still have the top error in `result`

stdweird:
@ellert wrt the return code, if i revert the last commit, things should be ok, no?
wrt the documentation: the current plugin does the following: if the proxy has an eppn extension, use it as username, if not, fallback to (usual?) gridmap with certificate DN (usual, well, it's not that it uses the same code as the regular gridmap plugin, so who knows).
so the current plugin would better be called `globus_eppn_or_gridmap`.

this patch adds the possibility to do the gridmap on the eppn from the attribute. we use this because we one one hand use cilogon proxies and on the other hand, we only have a mapping from eppn to local users to work with (and the cilogon DN does not allow such a mapping anyway, it ahs a random serial in it)

if this is clear, i can add this as comment to the 2nd commit

gsi/gridmap_eppn_callout/globus_gridmap_eppn.c
ellert: Calling GLOBUS_GRIDMAP_CALLOUT_ERROR with GLOBUS_SUCCESS as input is correct if the error you are reporting is the original error and isn't caused by another error.
Calling GLOBUS_GRIDMAP_CALLOUT_ERROR with GLOBUS_FAILURE as input will create a confusing error message stating that the error you are reporting was caused by a different unknown error - which is not true.

ellert:
> @ellert wrt the return code, if i revert the last commit, things should be ok, no?

Correct.

ellert:
Regarding documentation: the Globus gridmap ePPN callout package provides a plugin, not a library with an API to code against. This means that its sources do not have doxygen mark-up, since that only makes sense for packages that provide an API. So there is no documentation in the source code itself that needs to be updated with the new cuntionality.
That said, adding comments to the code explaining what it does is not a bad idea.

fscheiner:
@stdweird 
Can you add explaining comments to the code?

fscheiner:
@stdweird 
We're preparing a new release and the code in this PR could be included, but

* https://github.com/gridcf/gct/pull/73/commits/aabff9f7ae560e0091e38b68a745388f82913343 still needs to be reverted
* comments should be added to the code - or @ellert is it sufficient to add the comments to the actual commit message for https://github.com/gridcf/gct/pull/73/commits/e0d1f792a1b43cf423d69b933c82b6942668c4f2 ?

stdweird:
@fscheiner apologies, i completely forgot about this. will fix tomorrow

stdweird:
@fscheiner is https://github.com/gridcf/gct/pull/73/commits/6cc510d3794ec9c309515f15647e655dad8cd56a enough wrt documentation?

fscheiner:
> @fscheiner is [6cc510d](https://github.com/gridcf/gct/commit/6cc510d3794ec9c309515f15647e655dad8cd56a) enough wrt documentation?

Yes, thanks for updating the PR.

PR #74, opened 2019-04-05
Fix sed invocation on mac

matyasselmeci:
A user ran into this while trying to use the script on a Mac.


matyasselmeci:
I'm wrong about why it was broken, but this should fix the problem.

PR #76, opened 2019-04-12
Update documentation links to always point to the latest documentation

ellert:
The documentation link update mostly changed files in packaging/, but it did change two files in the sources. Bump versions for those.


fscheiner:
@ellert 
Thanks for anticipating this.

PR #78, opened 2019-04-18
Cleanup

ellert:
Here is a collection of clean-ups for the GCT sources. It is split in a series of commits to keep each change small.


prep-gsissh
```diff
@@ -1,7 +1,7 @@
 #! /bin/sh
 
 openssh_version=7.5p1
-gsissh_version=${openssh_version}b
+gsissh_version=${openssh_version}c
 openssh_tarball=openssh-${openssh_version}.tar.gz
 Source0=http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/${openssh_tarball}
```
fscheiner: @ellert 
If we're already at it, as there's no signature checking done, we should at least change to HTTPS, so the downloaded file cannot be modified in-flight unnoticed.

fscheiner: I tried to online-edit the file and use the "propose file change" functionality, but I always get a 404 page when I hit that button.

ellert: Changed...

PR #80, opened 2019-04-24
Fix openssl output format issue in sample-ca and gatekeeper startup script bug

iWangJiaxiang:
Hello guys,

I have encountered an error that `grid-ca-sign` can not find existing cert like below:

image

And when I checked the source code and  added some debug codes, I found there is a mistaken `sed` expression:

image
image

    sed -e "s|subject= *|/|" -e "s|,|/|g"

The cert subject in cert.pem begins with "subject= /O=...". However, the sed expression will process this line incorrectly with one more "/" before the result like "//O=...", which will never match the "${req_subj}" starting with "/O=...". So the `if` expression will never work.

And  I just simply delete the "/" and the new `sed` expression is as below:

    sed -e "s|subject= *||" -e "s|,|/|g"

Hope this will help!
Thank you.

### DIFF
--- a/gram/gatekeeper/source/init/globus-gatekeeper-lsb.in
+++ b/gram/gatekeeper/source/init/globus-gatekeeper-lsb.in
@@ -86,7 +86,7 @@ start()
         kflag=""
     fi
 
-    start_daemon -p "${GLOBUS_GATEKEEPER_PIDFILE}" \
+    start_daemon \
         ${GLOBUS_GATEKEEPER_NICE_LEVEL:+-n "${GLOBUS_GATEKEEPER_NICE_LEVEL}"} \
         "$prog" \
         -pidfile ${GLOBUS_GATEKEEPER_PIDFILE} \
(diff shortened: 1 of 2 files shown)


iWangJiaxiang:
Thank you! You solution is perfect and it works on my machines now~

iWangJiaxiang:
Hello,

The latest commit fixed an ancient bug on Redhat OS.

This bug only happens on Redhat OS, on which the compiler will use [globus-gatekeeper-lsb.in.](https://github.com/gridcf/gct/blob/master/gram/gatekeeper/source/init/globus-gatekeeper-lsb.in)  to build `ini.d` script. When I tried to start globus gatekeeper, the process was stuck forever. I added `set -x` code on the top of script and got

image

The "-p" pamameter should be used to defind port. however, there is incorrect use in the `ini.d` script for Redhat OS.

I removed the repeated as well as incorrect use of `-p` and everything works as expected! :D

image

Hope this would help.

PR #82, opened 2019-05-14
Keep peers in sync after SSL handshake failure in gssapi.

JaimeFrey:
In gss_init_sec_context(), if globus_i_gsi_gss_handshake() returns an
error, the code attempts to fetch the output_token (if anyy) twice.
First in gss_init_sec_context() and again in gss_delete_sec_context().
The second fetch clobbers any data from the first fetch, so the client
never sends it.
The fix is to not pass output_token to gss_delete_sec_context() if it
already contains data.

### DIFF
--- a/gsi/gssapi/source/library/init_sec_context.c
+++ b/gsi/gssapi/source/library/init_sec_context.c
@@ -554,7 +554,7 @@ GSS_CALLCONV gss_init_sec_context(
 
     gss_delete_sec_context(&local_minor_status, 
                            (gss_ctx_id_t *) &context,
-                           output_token);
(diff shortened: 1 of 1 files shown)


brianhlin:
@matyasselmeci the GRAM5 EL6 tests failed originally and so I re-ran it. Looks like the second time's the charm? Not sure why GitHub says it's failing -- perhaps because the original failure the deploy task didn't run at all and that's what's throwing it off.

JaimeFrey:
I ran the same tests interactively on EL6 with and without my patch, and both failed the same way as the automated test.

PR #83, opened 2019-05-17
Also use current tag in API documentation created by doxygen.

fscheiner:
This adds a HTML header for the HTML API documentation created by doxygen which gets the "current" tag (i.e. what `git describe --tags` prints out, see DESCRIPTION in `git-describe(1)` for details) inserted after the "Globus" version number as `(tag: <CURRENT_TAG>)`.

The top of the API docs will then look like this:

api-docs-header-with-current-tag-included

...when `HEAD` doesn't point to a tagged version or just show a tag (e.g. `v6.2.20190226`) after `(tag: ` otherwise.

This PR derives from the discussion in https://github.com/gridcf/gct-docs/pull/9.

****

@matyasselmeci @ellert @msalle 
Could you please have a look into my modifcations to `configure.ac` and `Makefile.am`. I'm actually not really familiar with that stuff and just tried to "imitate" what saw in the files.

Instead of the `sed` command in the Makefile we could maybe also just use `configure` to make the replacement, but I don't yet know, how to do that. I assume the file to be changed (with `.in` as suffix) needs to be included somewhere in `configure.ac` and contain an e.g. @VARIABLE@ instead. Any help appreciated, I can then force-push the changes.

****

There is one issue with my changes: The first time I issue `autoreconf -i` to create `configure` I get a non-zero exit value:

```
[johndoe@centos-7x-x86-64 gct]$ time autoreconf -i
[...]
+ cat
configure.ac:32: error: possibly undefined macro: m4_include
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/bin/autoconf failed with exit status: 1

real	7m50.942s
user	6m6.586s
sys	0m50.597s

[johndoe@centos-7x-x86-64 gct]$ echo $?
1

[johndoe@centos-7x-x86-64 gct]$ sed -n '32p' < configure.ac
m4_include([gsi_openssh/version.m4])
```

...but re-issuing the same command afterwards actually works:

```
[johndoe@centos-7x-x86-64 gct]$ time autoreconf -i
[...]
+ cat
configure.ac:43: installing 'build-aux/compile'
configure.ac:39: installing 'build-aux/config.guess'
configure.ac:39: installing 'build-aux/config.sub'
configure.ac:36: installing 'build-aux/install-sh'
configure.ac:36: installing 'build-aux/missing'
libltdl/Makefile.am: installing 'build-aux/depcomp'

real	7m5.515s
user	5m24.904s
sys	0m48.267s
[johndoe@centos-7x-x86-64 gct]$ echo $?
0
```

Without the modification (at 6dbb77120bf85d2972dcaa69f5dbf54bf5e7d260) it worked at the first try. I'm unsure why as I don't recognize the relation.

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -1,5 +1,5 @@
 PROJECT_NAME           = "Grid Community Toolkit"
-PROJECT_NUMBER         = @VERSION@
+PROJECT_NUMBER         = "@VERSION@ (tag: @GCT_CURRENT_TAG@)"
 OPTIMIZE_OUTPUT_FOR_C  = YES
 GENERATE_MAN           = YES
 MAN_LINKS              = YES
--- a/configure.ac
+++ b/configure.ac
@@ -1,3 +1,14 @@
+m4_define([gct_current_tag], [m4_syscmd([
+if test -e .git ; then
+    git describe --tags > gct-current-tag.inc.new;
+    if ! cmp gct-current-tag.inc.new gct-current-tag.inc > /dev/null 2>&1; then
+        mv gct-current-tag.inc.new gct-current-tag.inc
+    else
+        rm gct-current-tag.inc.new
+    fi
+fi])dnl
+m4_[]include(gct-current-tag.inc)])dnl
+
 m4_define([globus_buildno], [m4_syscmd([
 if test -e .git ; then
     git log -n 1 --pretty=format:%ct > globus-version.inc.new;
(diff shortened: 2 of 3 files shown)


fscheiner:
Ok, looks like Travis-CI didn't show the problem I mentioned above or just ignored it.

fscheiner:
@ellert 
> Instead of introducing a custom header, wouldn't it be simpler to just change the PROJECT_NUMBER in the Doxyfile.in. For consistency I would also add the tag in doxygen/index.dox.in.

Sure, that's the better solution. I'll modify my changes accordingly and force push the result.

> Running "sed -i" on a source file (doxygen/header.html) is bad.

I thought so, too, but didn't know better then,

> Better name the file with a .in suffix and add it to AC_CONFIG_FILES in configure.ac.

I see, this is how it's done. Thanks for the info.

fscheiner:
@ellert 
I now included your proposed changes and removed the unnecessary stuff from this PR. Building the API docs results in now:
api-docs-with-current-tag-included-simplified

****

OK to merge?

Issue #85, opened 2019-05-20
Change GSSAPITrustDns default in gsissh


DrDaveD:
The current default of GSSAPITrustDns to 'yes' in gsissh is historical and no longer needed.  Please change it to 'no'.

fscheiner:
Included in [GCT v6.2.20190906](https://github.com/gridcf/gct/releases/tag/v6.2.20190906).

PR #86, opened 2019-05-28
Change GSSAPITrustDNS default to no

ellert:
This PR was created in response to issue #85.


msalle:
Hi, I think the same should also be done for Fedora. Should we open a separate issue there?
(openssh-8.0p1-gssapi-keyex.patch adds "GSSAPITrustDNS no" but openssh-8.0p1-gsissh.patch changes the no into a yes)

ellert:
> Hi, I think the same should also be done for Fedora. Should we open a separate issue there?

No need to file an issue I think — I can update the Fedora patch once this PR is approved.

PR #89, opened 2019-06-10
Document install procedure

sfiligoi:
Add INSTALL.txt, that outlines how to build the software by hand.

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -5,11 +5,11 @@ Minimal tools needed
 --------------------
 
 The minimal RPMs (as provided by CentOS7 yum) needed to build this package are:
-autoconf automake make libtool libtool-ltdl gcc gcc-c++ patch openssl openssl-devel
+autoconf automake make libtool libtool-ltdl libtool-ltdl-devel gcc gcc-c++ patch openssl openssl-devel
 
 Additionally, if UDT support is desired, you will need
 epel-release
-udl
+udt
 
 Install instructions:
 --------------------
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -1,25 +1,34 @@
 Installation instructions
 =========================
 
-Minimal tools needed
---------------------
+Minimal tools needed (for CentOS7)
+---------------------------------
 
 The minimal RPMs (as provided by CentOS7 yum) needed to build this package are:
 autoconf automake make libtool libtool-ltdl libtool-ltdl-devel gcc gcc-c++ patch openssl openssl-devel
 
+If not already installed, please also add
+ curl
+
 Additionally, if UDT support is desired, you will need
-epel-release
-udt
+udt-devel
+
+Note: udt-devel comes from the EPEL repositories. Install 
+          epel-release 
+      to get them
 
 Install instructions:
 --------------------
 
 autoreconf -i
 ./configure <options>
+make -j <numcpus>
+make -j <numcpus>
+
 # for example:
 # ./configure --disable-gram5 --disable-gsi-openssh --enable-udt
-make -j 4
-make -j 4 install
+# make -j 4
+# make -j 4 install
 
 The files will be avaialble in
 /usr/local/globus-6/*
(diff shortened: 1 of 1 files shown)


brianhlin:
@matyasselmeci do these instructions look right to you?

INSTALL.txt
```diff
@@ -0,0 +1,27 @@
+Installation instructions
+=========================
+
+Minimal tools needed
+--------------------
+
+The minimal RPMs (as provided by CentOS7 yum) needed to build this package are:
+autoconf automake make libtool libtool-ltdl libtool-ltdl-devel gcc gcc-c++ patch openssl openssl-devel
```
matyasselmeci: At the moment, you also need curl because it's used for downloading the gsi-openssh patches. IIRC, this happens in the autoreconf command, i.e. before you get the chance to specify --disable-gsi-openssh.

Alternatively, you can tell people to delete the "prep-gsissh" script if they don't want to build gsi-openssh.

INSTALL.txt
```diff
@@ -0,0 +1,27 @@
+Installation instructions
+=========================
+
+Minimal tools needed
```
matyasselmeci: You might want to put up front that these instructions are for RHEL/CentOS/SL 7 and will need some modifications to work on Debian, for example. (We ought to write instructions for Debian and other distros eventually but one thing at a time I guess.)

INSTALL.txt
```diff
@@ -0,0 +1,27 @@
...
+--------------------
+
+The minimal RPMs (as provided by CentOS7 yum) needed to build this package are:
+autoconf automake make libtool libtool-ltdl libtool-ltdl-devel gcc gcc-c++ patch openssl openssl-devel
+
+Additionally, if UDT support is desired, you will need
+epel-release
+udt
```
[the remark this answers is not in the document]
sfiligoi: Pretty sure I did not need to install udt-release.

And, yes, udt comes from epel-release.
If you have a better way to express that, please let me know.

INSTALL.txt
```diff
@@ -0,0 +1,27 @@
...
+
+Install instructions:
+--------------------
+
+autoreconf -i
+./configure <options>
+# for example:
+# ./configure --disable-gram5 --disable-gsi-openssh --enable-udt
+make -j 4
```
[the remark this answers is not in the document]
sfiligoi: Correct.
But having to specify something (without is unlimited, which is undesirable), 4 is a good guess.

INSTALL.txt
sfiligoi: Pretty sure I did not need to install curl explicitly.
Maybe it came with the base CentOS7 install.
If you know it is needed, feel free to add it.

INSTALL.txt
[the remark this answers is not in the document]
matyasselmeci: When I tried it without udt-devel, the make died with
```
make[4]: Entering directory `/vagrant/xio/drivers/udt/source/udt'
${TAR-tar} -zxf udt.sdk.4.11.tar.gz

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
make[4]: *** [udt-unpack-stamp] Error 2
make[4]: Leaving directory `/vagrant/xio/drivers/udt/source/udt'
```
Installing udt-devel and rerunning configure and make solved the problem.

As for phrasing, how about something like
> If UDT support is desired, you will also need to install `udt-devel`.
>
> Note: udt-devel comes from the EPEL repositories. Install `epel-release` to get them.

sfiligoi:
I think I have addressed all the comments.

INSTALL.txt
```diff
@@ -0,0 +1,36 @@
...
+          epel-release 
+      to get them
+
+Install instructions:
+--------------------
+
+autoreconf -i
+./configure <options>
+make -j <numcpus>
+make -j <numcpus>
```
matyasselmeci:
```suggestion
make -j <numcpus> install
```
(looks like a duplicate line)

Issue #12, opened 2018-01-18
Replace toolkit.globus.org refs w/ refs to our own doc fork


matyasselmeci:
References to the original Globus documentation should be replaced to point to our documentation. From Mattias Ellert:

> There are lots of
> references to the toolkit.globus.org website and the Globus Alliance
> that need to be changed in the specfiles. For example, each package has
> a README file providing links to the globus documentation. I know the
> sources for the documentation were forked, but do we have a new
> location?
>
> == This is the README in the globus-common package ==
> $ cat /usr/share/doc/globus-common/README 
> This package is part of the C Common Libraries component
> of the Globus Toolkit. For more information visit:
>
> http://toolkit.globus.org/toolkit/docs/latest-stable/ccommonlib/
>
> Developer's Guide:
> http://toolkit.globus.org/toolkit/docs/latest-stable/ccommonlib/developer/
>
> Release Notes:
> http://toolkit.globus.org/toolkit/docs/latest-stable/ccommonlib/rn/
>
> Public Interface Guide:
> http://toolkit.globus.org/toolkit/docs/latest-stable/ccommonlib/pi/
> 
> Quality Profile:
> http://toolkit.globus.org/toolkit/docs/latest-stable/ccommonlib/qp/
>
> Migrating Guide:
> http://toolkit.globus.org/toolkit/docs/latest-stable/ccommonlib/mig/
> == End of file ==

ellert:
The README files are updated to point to the GCT documentation in PR #31:

This package is part of the C Common Libraries component
of the Grid Community Toolkit. For more information visit:

https://gridcf.org/gct-docs/ccommonlib/

Developer's Guide:
https://gridcf.org/gct-docs/ccommonlib/developer/

Release Notes:
https://gridcf.org/gct-docs/ccommonlib/rn/

Public Interface Guide:
https://gridcf.org/gct-docs/ccommonlib/pi/

Quality Profile:
https://gridcf.org/gct-docs/ccommonlib/qp/

Migrating Guide:
https://gridcf.org/gct-docs/ccommonlib/mig/

fscheiner:
This issue looks fixed to me, so I'm closing it.

msalle:
Not sure if any of these are still relevant, but there are still some references to old documentation at globus.org.

- [xio/drivers/gsi/globus_xio_gsi.h#L108](https://github.com/gridcf/gct/tree/master/xio/drivers/gsi/globus_xio_gsi.h#L108)
[...]

PR #92, opened 2019-07-03
Update additional old document references

fscheiner:
This fixes the remaining old document references from the list on https://github.com/gridcf/gct/issues/12#issuecomment-437045059.

****

Regarding the document from the old www-unix.globus.org site, looking into
https://web.archive.org/web/20100209082809/http://www-unix.globus.org/toolkit/docs/4.0/admin/docbook/ch05.html#s-prewsaa-env
https://gridcf.org/gct-docs/latest/gsic/pi/index.html#gsic-pi-env seems to
be the appropriate replacement.

### DIFF
--- a/gsi/cert_utils/source/programs/grid-cert-info.in
+++ b/gsi/cert_utils/source/programs/grid-cert-info.in
@@ -81,7 +81,7 @@ ${short_usage}
 EOF
 }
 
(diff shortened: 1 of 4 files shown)


fscheiner:
@matyasselmeci 
> Looks good. Sorry this took so long.

No issue at all. I usually don't manage to respond every day of the week, too.

PR #94, opened 2019-07-13
Updates

ellert:
* Replace broken URL

The link to the copy of the UDT source on toolkit.globus.org no longer works. Replace it with a link to the original on sourceforge.

* Remove LICENSE.globus file

In the myproxy sources, this file was relevant for the usage statistics collection code. This was removed, so the license file can be removed too.

* Do not package myproxy-server-setup / Reorganize documentation packaging

Make the Debian myproxy packaging similar to the Fedora/EPEL packaging.

* Add AC_CONFIG_MACRO_DIR and ACLOCAL_AMFLAGS to packages where missing

(except for noarch packages that do not need them)


matyasselmeci:
Looks like this broke the ssh builds on centos:centos6?

PR #96, opened 2019-07-16
Fix test in case of no sgids

msalle:
`getgroups()` can also return 0 or -1, these cases should be handled separately:
- fail in case getgroups returns -1
- use getgid() in case getgroups returns 0 (no secondary groups).

We also use `gid_t` instead of `uid_t` as type for my_groups.
Also note that `getgroups()` works differently on BSD and Linux.

### DIFF
--- a/gridftp/server/src/test/sharing_allowed_test.c
+++ b/gridftp/server/src/test/sharing_allowed_test.c
@@ -91,7 +91,8 @@ test_cleanup:
 int main()
 {
     uid_t                               my_uid;
-    uid_t                               my_groups[NGROUPS_MAX];
+    gid_t                               my_groups[NGROUPS_MAX];
+    gid_t                               my_default_gid;
     char *                              my_username;
     char *                              my_default_group;
     char *                              my_other_group;
@@ -124,10 +125,19 @@ int main()
     }
 
     rc = getgroups(NGROUPS_MAX, my_groups);
-    grent = getgrgid(my_groups[0]);
+    if (rc<0)
+    {
+        fprintf(stderr, "Error: getgroups(NGROUPS_MAX, my_groups) failed: %s\n",
+                strerror(errno));
+        exit(99);
+    }
+    my_default_gid = (rc>0 ? my_groups[0] : getgid());
+    grent = getgrgid(my_default_gid);
     if (grent == NULL)
     {
-        fprintf(stderr, "Unable to determine my default group name");
+        fprintf(stderr,
+                "Unable to determine my default group name (gid %d)\n",
+                my_default_gid);
         exit(99);
     }
     my_default_group = strdup(grent->gr_name);


ellert:
The test assumes that my_groups[0] is the user's primary group. This is not guaranteed. The test should always use getgid() in the primary group check. And then step through the array returned by getgroups until it finds a group id different from the primary group and use that fir the secondary group test (If one is found). This loop could also skip unknown groups. Something like this:
https://github.com/ellert/gct/commit/5a5fe8564559bb289870014b0e71658ae6b18dd4

ellert:
@LocutusOfBorg: Does this work for launchpad?
https://github.com/ellert/gct/commit/5a5fe8564559bb289870014b0e71658ae6b18dd4

msalle:
> The test assumes that my_groups[0] is the user's primary group. This is not guaranteed. The test should always use getgid() in the primary group check. And then step through the array returned by getgroups until it finds a group id different from the primary group and use that fir the secondary group test (If one is found). This loop could also skip unknown groups. Something like this:
> [ellert@5a5fe85](https://github.com/ellert/gct/commit/5a5fe8564559bb289870014b0e71658ae6b18dd4)

@ellert I agree your fix is probably the better one. Feel free to open a new pull request from your fix and close this one.
The problem is that the semantics of getgroups are a bit non-standard hence I tried to solve primaritly the most obvious problem. We also might want to skip the effective GID when different from the pGID and when present among the groups (whether it's included is unspecified, according to the Linux specific manpage).

LocutusOfBorg:
Hello, thanks I confirm it works with this patch! https://github.com/ellert/gct/commit/5a5fe8564559bb289870014b0e71658ae6b18dd4

LocutusOfBorg:
@ellert please fix this:


```
........./bin/bash ../libtool  --tag=CC   --mode=link gcc  -g -O2 -fdebug-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -dlpreopen force -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,--as-needed -Wl,-z,defs -o cmp_alias_ent_test cmp_alias_ent_test.o ../libglobus_gridftp_server.la  -lglobus_common -lglobus_xio -lglobus_gfork -lglobus_gridftp_server_control -lglobus_ftp_control -lglobus_authz -lglobus_gssapi_gsi -lglobus_gss_assist -lglobus_gsi_credential -lglobus_gsi_sysconfig -lglobus_io -lssl -lcrypto -lltdl 
..........sharing_allowed_test.c: In function ‘main’:
sharing_allowed_test.c:152:20: warning: passing argument 1 of ‘printf’ from incompatible pointer type [-Wincompatible-pointer-types]
  152 |             printf(stderr, "Found unknown group (gid %u). Ignoring.\n",
      |                    ^~~~~~
      |                    |
      |                    FILE * {aka struct _IO_FILE *}
In file included from /usr/include/stdio.h:867,
                 from sharing_allowed_test.c:1:
/usr/include/s390x-linux-gnu/bits/stdio2.h:105:32: note: expected ‘const char * restrict’ but argument is of type ‘FILE *’ {aka ‘struct _IO_FILE *’}
  105 | printf (const char *__restrict __fmt, ...)
      |         ~~~~~~~~~~~~~~~~~~~~~~~^~~~~
```

LocutusOfBorg:
it is an fprintf?

msalle:
> it is an fprintf?

He has already fixed it (based on my comment https://github.com/gridcf/gct/pull/99#pullrequestreview-264838994). See https://github.com/gridcf/gct/pull/99/files/ for the current patch.

LocutusOfBorg:
I confirm the #99 works thanks for giving me the full version

ellert:
Closed in favour of #99.

Issue #95, opened 2019-07-16
bad test checking for default group [patch]


LocutusOfBorg:
Hello, some build environments (e.g. launchpad builders) don't have a default group, so we have a failure in 
```

ERROR: sharing_allowed_test
===========================

Unable to determine my default group nameERROR sharing_allowed_test (exit status: 99)
```
https://launchpadlibrarian.net/433273991/buildlog_ubuntu-eoan-amd64.globus-gridftp-server_13.11-2_BUILDING.txt.gz

we patched the source to not fail on that test...
Is it possible to have some upstream workaround or to disable that test?
this is the "patch"
http://launchpadlibrarian.net/433274011/globus-gridftp-server_13.11-2_13.11-2ubuntu1.diff.gz
```patch
--- globus-gridftp-server-13.8.orig/test/sharing_allowed_test.c
+++ globus-gridftp-server-13.8/test/sharing_allowed_test.c
@@ -128,7 +128,7 @@ int main()
     if (grent == NULL)
     {
         fprintf(stderr, "Unable to determine my default group name");
-        exit(99);
+        exit(77);
     }
     my_default_group = strdup(grent->gr_name);
     if (my_default_group == NULL)
```

msalle:
Hi, I actually think there are two issues.
Firstly, there might not be any secondary gids, in which case `getgroups()` in [line 126](https://github.com/gridcf/gct/blob/master/gridftp/server/src/test/sharing_allowed_test.c#L126) will return 0 and `my_groups[0]` should not be used (it will probably be 0). It could also return a -1 on error, and also then `my_groups[0]` should not be used.
Secondly, there might be secondary gids, but the mapping from gid to a group name might not exist for various reasons. The latter case will give `grent == NULL`, your scenario.

LocutusOfBorg:
So, is it ok to patch this?

msalle:
I agree it needs fixing in the test. Probably by using getgid() instead of my_groups[0] in case rc==0 and producing a real error in case rc==-1.
I'll try to make a pull request, but it might take some days.

msalle:
Ok, here is https://github.com/gridcf/gct/pull/96
Assuming launchpad does actually provide groupnames (otherwise it's a launchpad bug I'd say), I think this should fix your issue.

LocutusOfBorg:
Looks like it didn't work...

```
umask 077; /usr/bin/openssl genrsa -out testcred.key 2048
:
[...]

PR #99, opened 2019-07-22
Avoid unknown secondary groups in test. Causes failures on launchpad.

ellert:
See discussion in #95 and #96.

### DIFF
--- a/gridftp/server/src/test/sharing_allowed_test.c
+++ b/gridftp/server/src/test/sharing_allowed_test.c
@@ -41,18 +41,25 @@ test_sharing_allowed(const sharing_allow_test_case_t *test_case)
     globus_l_gfs_data_session_t         session_handle = 
     {
         .username = test_case->sharing_user,
-        .gid_count = getgroups(0, NULL) + 1,
+        .gid_count = getgroups(0, NULL)
     };
 
-    session_handle.gid_array = malloc(session_handle.gid_count * sizeof(gid_t));
+    session_handle.gid_count =
+        (session_handle.gid_count < 0) ? 1 : session_handle.gid_count + 1;
+    session_handle.gid_array =
+        malloc(session_handle.gid_count * sizeof(gid_t));
+
     TEST_ASSERT(session_handle.gid_array != NULL);
 
-    rc = getgroups(session_handle.gid_count, session_handle.gid_array);
-    if (rc < session_handle.gid_count)
+    if (session_handle.gid_count > 1)
     {
-        session_handle.gid_array[session_handle.gid_count-1] = getegid();
+        session_handle.gid_count =
+	    getgroups(session_handle.gid_count, session_handle.gid_array);
+        session_handle.gid_count =
+            (session_handle.gid_count < 0) ? 1 : session_handle.gid_count + 1;
     }
 
+    session_handle.gid_array[session_handle.gid_count - 1] = getegid();
     session_handle.gid = getegid();
 
     if (test_case->sharing_users_allow)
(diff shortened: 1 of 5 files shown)


gridftp/server/src/test/sharing_allowed_test.c
```diff
@@ -137,19 +138,30 @@ int main()
...
+    rc = getgroups(NGROUPS_MAX, my_groups);
+    for (int i = 0; i < rc; i++)
+    {
+        if (my_groups[i] == my_gid)
+            continue;
+        grent = getgrgid(my_groups[i]);
+        /* Check for unknown groups (launchpad) */
+        if (grent == NULL)
+        {
+            printf(stderr, "Found unknown group (gid %u). Ignoring.\n",
```
msalle: typo: `printf` -> `fprintf`

gridftp/server/src/test/sharing_allowed_test.c
```diff
@@ -137,19 +138,30 @@ int main()
                 strerror(errno));
         exit(99);
     }
-    if (rc < 2)
+
+    grent = NULL;
+    rc = getgroups(NGROUPS_MAX, my_groups);
```
msalle: Perhaps good to also fail on `rc<0`

gridftp/server/src/test/sharing_allowed_test.c
```diff
@@ -123,8 +124,8 @@ int main()
         exit(99);
     }
 
-    rc = getgroups(NGROUPS_MAX, my_groups);
-    grent = getgrgid(my_groups[0]);
+    my_gid = getgid();
+    grent = getgrgid(my_gid);
     if (grent == NULL)
     {
         fprintf(stderr, "Unable to determine my default group name");
```
msalle: Probably good to log `my_gid` here, since it's odd it doesn't have a name, so could help with debugging.

ellert: As far as I understood the issue, the "unknown group" was never the primary one, but one of the secondary ones. The reason that the error was triggered for the primary group was that the test wrongfully assumed that my_groups[0] always contained the primary group. It was in reality containing a secondary group. The code logs when it encounters an unknown secondary group (a few lines further down). The test now uses getgid() to get the primary group.

msalle: It's true that it was one of the sgids, not the pgid, but the error was triggered because the sgid had no corresponding group, hence the getgrgid returned NULL. See https://github.com/gridcf/gct/issues/95#issuecomment-512180935: "Unable to determine my default group name (gid 110)" and my reply there https://github.com/gridcf/gct/issues/95#issuecomment-512265660, linking to a launchpad bug. We found this because it now logged the failing gid, hence my suggestion to add that in the fprintf in line 130 (131 after your commits).

gridftp/server/src/test/sharing_allowed_test.c
ellert: The result from getgroups() is not checked for negative values anywhere in the code.
See: gridftp/server/src/globus_i_gfs_data.c

If -1 is returned in the test code, the loop will be skipped, skip_other_group will be set to true and the secondary group tests will be skipped.

ellert: Actually, I think the code was written so that getgroups can not return -1. According to the man page, there are two error conditions:

       EFAULT list has an invalid address.
       EINVAL size is less than the number of supplementary group IDs, but is not zero.

These error conditions are checked in the code before calling getgroups.

msalle: That's indeed true for Linux (and MacOS if we can believe the man page), but not per se for other OSes. We can leave it as is, but personally I find it good practice to check for errors even if with the current code, they cannot appear.

LocutusOfBorg:
This one works for me! thanks!

Issue #62, opened 2018-11-29
Does -dcpriv work when talking to dCache?


onnozweers:
Globus Toolkit has a security vulnerability: when `globs-url-copy -dcpriv` is specified, and connecting to a dCache instance (which does not support data channel privacy), globus-url-copy silently decides to send the data in plain text anyway. The user is not warned and may assume the data is transferred privately while in fact it is not. 

A bug report is here: https://github.com/globus/globus-toolkit/issues/121

Is this bug present in GCT too?

Kind regards,
Onno

msalle:
Hi Onno,
the short answer is yes, since the gct version of globus-url-copy is currently fully backwards compatible with the globus-toolkit 6.0 version. We had a long discussion about the issue in the spring but there is no easy way to fix it in ways that will not unexpectedly break things, and furthermore, would require simultaneous changes in multiple clients (also uberftp for example) and servers. That basically where the discussion stalled... Perhaps others (such as Paul) have a good idea on how to got forward.

onnozweers:
Hi msalle, nice to meet you here! :-)

I'm very curious why it is so difficult to fix this. I would think, it's better to break some things than to secretly expose private data. But perhaps I'm overlooking some complications that make it more difficult than such a simple choice.

I'd like to add to the discussion that since this issue was reported, the GDPR has become effective. So perhaps with this bug, people might be (unknowingly) violating the GDPR. But I'm not a law expert.

Kind regards,
Onno

paulmillar:
I don't think this is really that difficult.

There are only two ways to fix this (that make sense):
1. fix `-dcpriv` so transfers fail if they are not encrypted.
2. make `-dcpriv` a no-op.  Document that it does nothing and it only remains for backwards compatibility.  Introduce a new option that enforces encryption.

The decision between 1. and 2. is choosing whether it is better to fail people who rely on `-dcpriv` not working correctly (who include it by mistake), or fail people who rely on `-dcpriv` working correctly (who didn't change the option).

Put another way, choosing option 1. will (possibly) result in failed transfers that should succeed; choosing option 2. will (possibly) result in successful transfers that should have failed.

From a security point-of-view, only 1. makes sense, because 2. opens a fresh security problem: people who didn't update to the new option.

onnozweers:
Nothing seems to happen. 😭

Could this have a higher urgency please?

fscheiner:
@onnozweers @paulmillar 
[...]

PR #100, opened 2019-07-25
Document guc/dcpriv behaviour with servers not supporting DC encryption

fscheiner:
Until we find a way to "fix" the problem described in #62 w/o breaking other functionality in the GridFTP server and client(s) this PR puts a warning into the guc manpage which - together with the apparently known fact (as per #69) that dCache does not support data channel encryption - hopefully prevents user from falling into that trap.

****
@ellert 
* I changed the text version of the manpage and the manpage directly, is there a tool I could use instead?
* Does this change also require a version number change for the gass-copy packages (in the spec and debian packaging files)? I'd say yes, but am unsure. If you agree, I can add these changes to this PR.

### DIFF
--- a/gass/copy/source/configure.ac
+++ b/gass/copy/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gass_copy],[10.4],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_gass_copy],[10.5],[https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
 AC_SUBST([MAJOR_VERSION], [${PACKAGE_VERSION%%.*}])
 AC_SUBST([MINOR_VERSION], [${PACKAGE_VERSION##*.}])
--- a/gass/copy/source/globus-url-copy.1
+++ b/gass/copy/source/globus-url-copy.1
@@ -1,13 +1,13 @@
 '\" t
 .\"     Title: globus-url-copy
 .\"    Author: [see the "AUTHOR" section]
-.\" Generator: DocBook XSL Stylesheets vsnapshot <http://docbook.sf.net/>
-.\"      Date: 03/31/2018
+.\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
+.\"      Date: 07/25/2019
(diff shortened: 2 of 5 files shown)


fscheiner:
If accepted, I will also update the documentation in https://github.com/gridcf/gct-docs.

ellert:
> @ellert
> 
>   * I changed the text version of the manpage and the manpage directly, is there a tool I could use instead?
> 
>   * Does this change also require a version number change for the gass-copy packages (in the spec and debian packaging files)? I'd say yes, but am unsure. If you agree, I can add these changes to this PR.

The rules to generate the man pages from the text sources are in the Makefiles. If configure is run with a2x or a2x.py installed the rules will be activated and you can jusr run "make globus-url-copy.1". a2x is provided by the asciidoc package.

See configure.ac and Makefile.am for details.

Yes, the version should be bumped.

fscheiner:
> @ellert
> The rules to generate the man pages from the text sources are in the Makefiles. If configure is run with a2x or a2x.py installed the rules will be activated and you can jusr run "make globus-url-copy.1". a2x is provided by the asciidoc package.
> 
> See configure.ac and Makefile.am for details.

Thanks for the clarification.

> Yes, the version should be bumped.

Ok, will do so.

fscheiner:
Took me a while to get everything to work on Armbian Stretch (based on Debian Stretch). Maybe useful for documentation: in addition to `build-essential` (although this might be a little overkill) one needs to install the following packages:
```
asciidoc libtool libltdl-devpkg-config docbook-xml docbook-xsl xsltproc
```
...to make this work.

****

BTW, that (`autoreconf -i && ./configure`) is a considerable effort for just changing a manpage. Why do we actually keep the resulting manpages in the repo if they're (re)created during a make run anyhow?

fscheiner:
Nearly forgot to ask: Is the version bump ok as is (`10.4` to `10.5`), or should I just bump the release (`10.4-1` to `10.4-2`)?

ellert:
> Nearly forgot to ask: Is the version bump ok as is (`10.4` to `10.5`), or should I just bump the release (`10.4-1` to `10.4-2`)?

You have done it correctly. This changes files that are in the source tarball, so the tarball name must change (globus_gass_copy-10.4.tar.gz → globus_gass_copy-10.5.tar.gz). Bumping the release keeping the version is for making a package update using the same source tarball, possibly adding patches to the source package.

PR #101, opened 2019-07-30
Fix test race condition

ellert:


matyasselmeci:
So the race condition is running the test before the server is up? Is there a better way of verifying that the server finished starting up?

msalle:
> So the race condition is running the test before the server is up? Is there a better way of verifying that the server finished starting up?

Assuming you have nc (netcat) you can do something like `nc -z localhost $port` but that won't give you output. Alternatively (also with nc) you can write a loop with an `nc -q1 -w1 $host $port < /dev/null` and then you could grep for some string that would be returned once the service is running. Without `nc` you can probably do some magic with bash's TCP socket support.

matyasselmeci:
That seems kinda complicated; for unit tests, I'm not sure if it's worth the effort. I'll approve this.

Issue #102, opened 2019-08-05
gridftp failures between ipv4 client and dual-stack server


okeeble:
We have seen gridftp failures on sites with a dual-stack server and ipv4-only clients. In certain configurations (e.g. the new Dirac default), the client will use EPSV to get an IP address for the data connection. This returns an ipv6 address which, in a simple upload/download scenario, is unusable by the client.

While this behaviour is understandable (the client may be trying to establish a third-party transfer between two ipv6 capable servers) in reality it's causing problems.

We will provide a PR to give the option for EPSV to return an ipv4 address over an ipv4 connection.

Related DPM ticket : https://its.cern.ch/jira/browse/LCGDM-2817

fscheiner:
Included in [GCT v6.2.20190906](https://github.com/gridcf/gct/releases/tag/v6.2.20190906).

PR #103, opened 2019-08-05
Patch for Issue #102

okeeble:
Patch supplied by Andrey Kiryanov for issue #102, "Gridftp failures between ipv4 client and dual-stack server".

This has been installed on a couple of DPM production sites for over a month.

### DIFF
--- a/common/source/configure.ac
+++ b/common/source/configure.ac
@@ -1,5 +1,5 @@
 AC_PREREQ([2.60])
-AC_INIT([globus_common], [18.4], [https://github.com/gridcf/gct/issues])
+AC_INIT([globus_common], [18.5], [https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
 AC_SUBST([MAJOR_VERSION], [${PACKAGE_VERSION%%.*}])
 AC_SUBST([MINOR_VERSION], [${PACKAGE_VERSION##*.}])
--- a/common/source/scripts/Paths.pm.in
+++ b/common/source/scripts/Paths.pm.in
@@ -14,6 +14,8 @@
 # 
 package Globus::Core::Paths;
 
+use File::Basename qw(dirname);
+
 require Exporter;
 
 @ISA = qw(Exporter);
@@ -66,6 +68,6 @@ $libexecdir = "@libexecdir@";
 $sysconfdir = "@sysconfdir@";
 $sharedstatedir = "@sharedstatedir@";
 $localstatedir = "@localstatedir@";
-$perlmoduledir = "@perlmoduledir@";
+$perlmoduledir = dirname(dirname(dirname(__FILE__)));
 
 1;
--- a/common/source/scripts/globus-hostname.in
+++ b/common/source/scripts/globus-hostname.in
@@ -40,7 +40,6 @@ libexecdir="@libexecdir@"
 sysconfdir="@sysconfdir@"
 sharedstatedir="@sharedstatedir@"
 localstatedir="@localstatedir@"
-perlmoduledir="@perlmoduledir@"
 
 PATH=${bindir}:${sbindir}:${PATH}
 
--- a/gass/copy/source/configure.ac
+++ b/gass/copy/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gass_copy],[10.4],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_gass_copy],[10.5],[https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
 AC_SUBST([MAJOR_VERSION], [${PACKAGE_VERSION%%.*}])
 AC_SUBST([MINOR_VERSION], [${PACKAGE_VERSION##*.}])
--- a/gass/copy/source/globus-url-copy.1
+++ b/gass/copy/source/globus-url-copy.1
@@ -1,13 +1,13 @@
 '\" t
 .\"     Title: globus-url-copy
 .\"    Author: [see the "AUTHOR" section]
-.\" Generator: DocBook XSL Stylesheets vsnapshot <http://docbook.sf.net/>
-.\"      Date: 03/31/2018
+.\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
+.\"      Date: 07/25/2019
 .\"    Manual: Grid Community Toolkit Manual
 .\"    Source: Grid Community Toolkit 6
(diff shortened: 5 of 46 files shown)


fscheiner:
I created a patch that also updates the `globus-gridftp-server(8)` manpage and the packaging information accordingly.

@ellert 
I bumped the minor version only, but maybe a bump of the major and minor version would be more appropriate, as functionality has changed (incl. globus-gridftp-server option)? Should we also change the minimum version numbers for the requirements?

****

Patch is attached. BTW, is there an easier way, e.g. like adding that patch as commit to Oliver's branch, just like I could rebase his branch earlier?

[updates-to-manpage-and-packaging-metadata.patch.txt](https://github.com/gridcf/gct/files/3482585/updates-to-manpage-and-packaging-metadata.patch.txt)

fscheiner:
I'd rephrase the commit message to something like:

```
Fix problems between dual-stack (IPv4/IPv6) servers and IPv4-only clients

Fixes #102 
```

After merging this should still automatically close #102, but the "important" information is in the first line.

fscheiner:
Just spotted [this comment in the CERN JIRA](https://its.cern.ch/jira/browse/LCGDM-2817?focusedCommentId=2681195&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2681195):

```
[...]
Note that one side-effect of this patch is that an ipv4 client will not be able
to establish an ipv6 transfer between a patched DPM and another ipv6
capable endpoint.
```

@ellert @matyasselmeci @msalle @okeeble 
Did the above described operation work in the past - i.e. control channels established via IPv4 and data channels established via IPv6 IIUIC? Does it constitute a problem when it no longer works with the patch included?

...ok, it's not enabled by default, so maybe not an issue.

matyasselmeci:
Does this change conflict with the epsv_ip option ("Adds an IPv6 address to EPSV response. Breaks RFC 2428, but allows redirection to work with IPv6.")? We have that option set by default for OSG sites that use gridftp (and I was hoping to make that the default in GCT if it doesn't cause any problems).

okeeble:
Thanks for taking a look at this.
@fscheiner - I don't know if this scenario ever happened in reality. With this patch the transfer would be established over ipv4, so things should not break.
@matyasselmeci - no, there's no conflict, but one could say this patch extends the scope of epsv_ip. It's the return of an address that breaks RFC 2428, and now we're allowing ourselves to return an ipv4 address under certain circumstances.

ellert:
> I bumped the minor version only, but maybe a bump of the major and minor version would be more appropriate, as functionality has changed (incl. globus-gridftp-server option)? Should we also change the minimum version numbers for the requirements?

If the change adds new functions to the API, the major version AND age should both be increased by one (which keeps the soname unchanged), and the minor version reset to 0. If there are dependent packages that use the additions to the API their (build)requirements should be increased accordingly (in configure.ac, *.spec and debian/control). The package where the major version changes should also change its debian/*.shlibs.

globus-gridftp-server-control must increase the major version. For globus-gridftp-server it is not necessary.

> Patch is attached. BTW, is there an easier way, e.g. like adding that patch as commit to Oliver's branch, just like I could rebase his branch earlier?

If you read carefully it says "Add more commits by pushing to the master branch on okeeble/gct."

okeeble:
@ellert - what we've done is add a runtime configuration option, which required adding one new function to globus_gridftp_server_control. Its API has therefore indeed changed, but in reality this is internal and noone needs to change any code. I just want to check that this is understood before we bump a major version.

ellert:
Bumping the major version together with the age means that the new library keeps the same soname and is backwards compatible. Anything that was linked to the previous version continues to work with the new version without recompiling or relinking. What the change in major version indicates is that everything linked to the new version might not work with the old version, which is true if it uses the new function.

We are not bumping the major version and resetting age to 0 here, which would be the right thing to do if we REMOVED something from the API or changed it in some backward incompatible way. In that case the soname changes and everything needs to be recompiled/relinked.

okeeble:
Thanks - makes perfect sense.

fscheiner:
@ellert 
Sorry for the delay, I was busy with other things last week.

> > I bumped the minor version only, but maybe a bump of the major and minor version would be more appropriate, as functionality has changed (incl. globus-gridftp-server option)? Should we also change the minimum version numbers for the requirements?
> 
> If the change adds new functions to the API, the major version AND age should both be increased by one (which keeps the soname unchanged), and the minor version reset to 0. If there are dependent packages that use the additions to the API their (build)requirements should be increased accordingly (in configure.ac, *.spec and debian/control). The package where the major version changes should also change its debian/*.shlibs.
> 
> globus-gridftp-server-control must increase the major version. For globus-gridftp-server it is not necessary.

Ok, thanks for the explanation. I hope I did it correctly now.

> > Patch is attached. BTW, is there an easier way, e.g. like adding that patch as commit to Oliver's branch, just like I could rebase his branch earlier?
> 
> If you read carefully it says "Add more commits by pushing to the master branch on okeeble/gct."

I see. Before I try to push my modified patch, can you please have a look if my library version changes and changes to the packaging metadata are correct now?

[updates-to-manpage-and-packaging-metadata-v2.patch.txt](https://github.com/gridcf/gct/files/3516617/updates-to-manpage-and-packaging-metadata-v2.patch.txt)

If they're correct, I'll then try to push directly to the master branch of Oliver's fork of the GCT.

ellert:
> I see. Before I try to push my modified patch, can you please have a look if my library version changes and changes to the packaging metadata are correct now?
> 
> [updates-to-manpage-and-packaging-metadata-v2.patch.txt](https://github.com/gridcf/gct/files/3516617/updates-to-manpage-and-packaging-metadata-v2.patch.txt)
> 
> If they're correct, I'll then try to push directly to the master branch of Oliver's fork of the GCT.

This looks OK.
If I run `globus-gridftp-server -help -asciidoc > globus-gridftp-server-options.txt`, the resulting file has 4 additional spaces on the line after "\*-epsv-match\*::" that are not present if I apply your patch. It is maybe not important - the manpage doesn't change due to those spaces.

fscheiner:
@ellert 
> > If they're correct, I'll then try to push directly to the master branch of Oliver's fork of the GCT.
> 
> This looks OK.
> If I run `globus-gridftp-server -help -asciidoc > globus-gridftp-server-options.txt`, the resulting file has 4 additional spaces on the line after "*-epsv-match*::" that are not present if I apply your patch. It is maybe not important - the manpage doesn't change due to those spaces.

Ok, I see, thanks for checking, I now use the output of `globus-gridftp-server [...]` instead of editing the file directly (my editor didn't highlight white space so I didn't notice).

****

Will merge the PR when Travis is finished.

PR #104, opened 2019-08-31
Add missing dependency

ellert:


fscheiner:
Should we also create a new candidate tag? For the API, we could then just replace the one merged yesterday with the new one maybe.

Issue #71, opened 2019-03-13
6.2.20190226 release in EPEL updates-testing?


jbasney:
Passing along a question I received from Derek Simmel:
> Jim,
> 
> I was trying to install various parts from the current gct release today...
> 
> The webpage at https://github.com/gridcf/gct/releases/tag/v6.2.20190226 says that precompiled packages are available in the EPEL updates-testing repo:
> 
> • EPEL (for Red Hat Enterprise Linux, CentOS and Scientific Linux 6 and 7 (currently in EPEL updates-testing!))
> 
> but I can't seem to find where one gets the updates-testing repo from... the epel-testing repo does not have the most current packages in it.
> 
> Can you please direct me to where I may find the updates-testing repo?
> 
> Thanks -
> 
> - Derek

fscheiner:
I believe the testing repo can be activated in the repo definitions in `/etc/yum.repos.d`. For example the [EPEL6 RPM](https://dl.fedoraproject.org/pub/epel/epel-release-latest-6.noarch.rpm) shows a file named `epel-testing.repo` where all repos are deactivated by default (`enabled=0`):

epel-testing.repo:
```
[epel-testing]
name=Extra Packages for Enterprise Linux 6 - Testing - $basearch
#baseurl=http://download.fedoraproject.org/pub/epel/testing/6/$basearch
mirrorlist=https://mirrors.fedoraproject.org/metalink?repo=testing-epel6&arch=$basearch
failovermethod=priority
enabled=0
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-6

[epel-testing-debuginfo]
name=Extra Packages for Enterprise Linux 6 - Testing - $basearch - Debug
#baseurl=http://download.fedoraproject.org/pub/epel/testing/6/$basearch/debug
mirrorlist=https://mirrors.fedoraproject.org/metalink?repo=testing-debug-epel6&arch=$basearch
failovermethod=priority
enabled=0
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-6
gpgcheck=1

[epel-testing-source]
name=Extra Packages for Enterprise Linux 6 - Testing - $basearch - Source
#baseurl=http://download.fedoraproject.org/pub/epel/testing/6/SRPMS
mirrorlist=https://mirrors.fedoraproject.org/metalink?repo=testing-source-epel6&arch=$basearch
failovermethod=priority
enabled=0
[...]

PR #105, opened 2019-09-13
Disallow TLS 1.0/1.1 by default again

ellert:
Following @maarten-litmaath's comment on #55, that the setting for what TLS versions are enabled by default could now be restored to only allow 1.2, I have created this PR to do that.

A similar question. What about the change in #19. Can that be reverted too now? Or is it still relevant?

### DIFF
--- a/gsi/gssapi/source/configure.ac
+++ b/gsi/gssapi/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gssapi_gsi],[14.11],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_gssapi_gsi],[14.12],[https://github.com/gridcf/gct/issues])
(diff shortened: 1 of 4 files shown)


maarten-litmaath:
Hi,
I do not know anything about #19 and propose we do not touch that area at this time.

fscheiner:
> A similar question. What about the change in #19. Can that be reverted too now? Or is it still relevant?

[SSL_CTX_set_options(3)](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_options.html) states that this option only affects CBC ciphers. So changing the cipher to something non-CBC for Bestman maybe could allow us to revert #71?

OTOH, the above manpage also states that this does only affect the SSL 3.0/TLS 1.0 protocol (assuming it was fixed in later protocol versions). And with the change in this PR we default to TLS 1.2 so `SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS` wouldn't apply anyhow to the defaults.

Issue #118, opened 2020-01-20
gsi-openssh silently fails to install if not running as root


Phidica:
I'm building the latest master on Debian because I wanted to get gsi-openssh working.

After running the following,
```
$ autoreconf --install --no-recursive
$ ./configure --prefix=$PWD/install --enable-gsi-openssh
$ make -j3
$ make -j3 install
```
I did not see `gsissh` in the ./install/bin directory. Note that I did not execute any of these commands with `sudo` because my install prefix is a local user directory and hence *it should not be necessary*.

I spotted the issue buried in the output of the `make install` run:
```
<trim>
./mkinstalldirs /home/phidica/Code/gct/install/share/man
./mkinstalldirs /home/phidica/Code/gct/install/share/man/man1
./mkinstalldirs /home/phidica/Code/gct/install/share/man/man5
./mkinstalldirs /home/phidica/Code/gct/install/share/man/man8
./mkinstalldirs /home/phidica/Code/gct/install/libexec
(umask 022 ; ./mkinstalldirs /var/empty)
mkdir /var/empty
mkdir: cannot create directory ‘/var/empty’: Permission denied
make[4]: *** [Makefile:346: install-files] Error 1
make[4]: Leaving directory '/home/phidica/Code/gct/gsi_openssh/source'
make[3]: *** [Makefile:2743: gsi-openssh-install] Error 2
<trim>
```

It appears that there is some call in the gsi-openssh install process that does not respect the install prefix I overrode with `./configure`, and the install fails because it attempts to create `/var/empty` as a normal user.

I solved the problem by running `sudo mkdir /var/empty` and repeating the install.

I have two independent interpretations of what the "issue" to be solved here is:
1. The gsi-openssh install process neglects to incorporate `$prefix` for this one directory, so it should instead attempt to create, eg, `$prefix/var/empty`.
1. The failure of this component was silent and no advice was provided on how to resolve the problem, eg, bring the install to a halt and print "Couldn't create /var/empty. Do you lack superuser permissions?"

I believe that implementing either one of these, individually, would resolve this issue.

msalle:
Hi,
thank you for opening the issue and providing feedback.
Several remarks. First of all, this is actually an issue with [openssh-portable](https://github.com/openssh/openssh-portable/) itself, inherited by 'our gct' version, so perhaps we should discuss it there (see reporting bugs in their README).
[...]

PR #119, opened 2020-01-28
On errors let make return with non-zero exitvalue

msalle:
This fixes https://github.com/gridcf/gct/issues/118
There are probably a few other places where we should bail out with a non-zero
exit value but this certainly fixes the main ones.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,9 +16,8 @@ $(GT_ALL_TARGETS_INSTALL):
 $(GT_TARGETS_CHECK):
 	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) check || exit 1
 $(GT_ALL_TARGETS_DIST):
-	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && MAKEFLAGS= $(MAKE) $(AM_MAKEFLAGS) dist && \
-	$(MKDIR_P) package-output && \
-	cp $($@_SUBDIR)/$($@-tarball) package-output || exit 1
+	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && MAKEFLAGS= $(MAKE) $(AM_MAKEFLAGS) dist || exit 1
+	@$(MKDIR_P) package-output && cp $($@_SUBDIR)/$($@-tarball) package-output || exit 1
 
 $(GT_ALL_STAMPS):
 	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; ( (cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) all) && touch $@) || exit 1
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,23 +8,23 @@ DIST_SUBDIRS = @TOP_DIST_SUBDIRS@
 
 $(GT_TARGETS_CLEAN):
 	@if [ -f $($@_stamp) ]; then \
-		export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) clean ; \
+		export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) clean || exit 1 ; \
 	fi
 	@rm -f $($@_stamp)
 $(GT_ALL_TARGETS_INSTALL):
-	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) install
+	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) install || exit 1
 $(GT_TARGETS_CHECK):
-	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) check
+	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) check || exit 1
 $(GT_ALL_TARGETS_DIST):
-	@$(MKDIR_P) package-output
-	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && MAKEFLAGS= $(MAKE) $(AM_MAKEFLAGS) dist
-	@cp $($@_SUBDIR)/$($@-tarball) package-output
+	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && MAKEFLAGS= $(MAKE) $(AM_MAKEFLAGS) dist && \
+	@$(MKDIR_P) package-output && \
+	@cp $($@_SUBDIR)/$($@-tarball) package-output || exit 1
 
 $(GT_ALL_STAMPS):
 	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; ( (cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) all) && touch $@) || exit 1
 
 $(GT_ONLY):
-	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) all
+	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) all || exit 1
 
 RPMDIRS = BUILD BUILDROOT RPMS SOURCES SPECS SRPMS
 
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -8,23 +8,23 @@ DIST_SUBDIRS = @TOP_DIST_SUBDIRS@
...
 $(GT_TARGETS_CHECK):
-	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) check
+	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) check || exit 1
 $(GT_ALL_TARGETS_DIST):
-	@$(MKDIR_P) package-output
-	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && MAKEFLAGS= $(MAKE) $(AM_MAKEFLAGS) dist
-	@cp $($@_SUBDIR)/$($@-tarball) package-output
+	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && MAKEFLAGS= $(MAKE) $(AM_MAKEFLAGS) dist && \
+	@$(MKDIR_P) package-output && \
+	@cp $($@_SUBDIR)/$($@-tarball) package-output || exit 1
```
fscheiner: @msalle:
Comparing the original code and the modification, `mkdir` was originally done in another dir than `$($@_SUBDIR)` (assuming that `$($@_SUBDIR)` != `$PWD`), but with the modification it's invoked in `$($@_SUBDIR)` - unless that `MAKEFLAGS= $(MAKE) $(AM_MAKEFLAGS) dist` changes the `$PWD`.

fscheiner: @msalle 
I retriggered the ["deploy" job](https://travis-ci.org/gridcf/gct/jobs/642843741) and it errors out the same way as during the original run, hence I tend to assume this error could be related to the changes and not an intermittent issue with Travis CI:

```
[...]
make[1]: Leaving directory `/gct/common/source'

/bin/sh: line 1: @/usr/bin/mkdir: No such file or directory

make: *** [globus_common-dist] Error 1

+ rm -f /tmp/tmp.FkzDiIlIxc

The command "travis-ci/setup_tasks.sh" exited with 2.
```

OTOH, when checking on my local Ubuntu 19.04 `mkdir` is not in `/usr/bin` but still in `/bin` - i.e. `/bin` and `/sbin` aren't yet "merged" with `/usr/bin` and `/usr/sbin` as on Fedora. Not sure why this should break now, though. :-/

matyasselmeci: You need to drop the leading `@` from the new `$(MKDIR_P)` line and the `cp` line below it since they're now part of the same shell invocation (due to the `&& \`) at the end of the previous lines. The error is not that it's not finding `mkdir` in `/usr/bin`, the error is that it's looking for a file named `@/usr/bin/mkdir`.

msalle: Good point, thanks. It should be fixed now.
We should at some point further cleanup the Makefile.

fscheiner: @matyasselmeci 
> You need to drop the leading `@` from the new `$(MKDIR_P)` line and the `cp` line below it since they're now part of the same shell invocation (due to the `&& \`) at the end of the previous lines. The error is not that it's not finding `mkdir` in `/usr/bin`, the error is that it's looking for a file named `@/usr/bin/mkdir`.

Thanks for the clarification. Actually that even was my initial idea: that some "unexpanded" command is fed to the shell. But then I thought why should the `@` make keyword loose meaning in a command list? But yeah, the manual tells (from https://www.gnu.org/software/make/manual/make.html#Recipe-Echoing):

```
[...]
When a line starts with ‘@’, the echoing of that line is suppressed.
The ‘@’ is discarded before the line is passed to the shell. [...]
```
...so it's only discarded at the start of the line. I'd then actually write it like that instead:

```
@ command arg1 arg2 [...]
```

...then it could be understandable from the context that `@` does not affect the command, but the whole line - but I could also RTFM instead, of course. ;-)

Makefile.am
```diff
@@ -8,23 +8,23 @@ DIST_SUBDIRS = @TOP_DIST_SUBDIRS@
...
 $(GT_TARGETS_CHECK):
-	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) check
+	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) check || exit 1
 $(GT_ALL_TARGETS_DIST):
-	@$(MKDIR_P) package-output
-	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && MAKEFLAGS= $(MAKE) $(AM_MAKEFLAGS) dist
-	@cp $($@_SUBDIR)/$($@-tarball) package-output
+	@export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && MAKEFLAGS= $(MAKE) $(AM_MAKEFLAGS) dist && \
+	$(MKDIR_P) package-output && \
+	cp $($@_SUBDIR)/$($@-tarball) package-output || exit 1
```
fscheiner: @msalle 
Resolved that too early.

The order of the commands is still different to the original code:

As the `cd in the first part (originally in the second line) changes `$PWD`, the `mkdir` in the second part (originally in the first line) is done in a different directory.

Why change the order?

msalle: I prefer to do the mkdir only after the make, since it doesn't leave empty directories in case of failures in the make process, but you're right something is still odd there, however the original is also wrong I think: it's not doing a `cd ..`, so how can the cp succeed, since we're still in the `$@_SUBDIR`, or is there some magic going on that I don't see?.
Also, how sure are we that `$@_SUBDIR` is only one level deeper? If so, we can just add a `cd ..` before the mkdir. 
I cannot really experiment right now, probably tomorrow. If you have a chance to figure this one out, I'll push an update (or you can just make a new PR instead).

fscheiner: Sorry, no time tomorrow, but can have a look into that next week.

fscheiner: @msalle:

Short update:
I seemed to remember having dealt with `make dist` in the past and just had a look into the scripts @matyasselmeci created for Travis CI. Some time ago I added a change to also create the full source installer tarball for the GCT and looking into https://github.com/gridcf/gct/blob/master/travis-ci/make_source_tarballs.sh#L42..L45, the `cp` operation in the Makefile indeed doesn't work as expected, because otherwise the source installer tarball wouldn't have to be manually copied to the `package-output` dir in the shell script.

msalle: Ok, I figured it out I think. The make automatically jumps back to the top_srcdir. I've now split the setup in two, and let both of them error out.
E.g. `make globus_common-dist` now correctly works again.
Sorry for all the mess )-:

fscheiner:
@msalle , @matyasselmeci :
I just wondered, if there maybe is a switch for `make` to cancel further processing of a makefile if a rule (or recipe line) exits with `!0`?
I mean that sounds like a legitimate case and for the "opposite" case - i.e. trying to "make" as many targets as possible to e.g. detect independent problems - there exists `-k|--keep-going`.

If there isn't a switch for our case, let's pull these changes.

matyasselmeci:
I'm confused. Exiting when a command returns nonzero is the default behavior for make.

I see why the `|| exit 1` is needed in the `@if [ -f $(@_stamp) ];`... line -- the `if` swallows up the exit code.  I don't understand why it's needed in any of the `$(MAKE) $(AM_MAKEFLAGS)`... lines -- if the `$(MAKE)` returns nonzero then that line already fails.

matyasselmeci:
Also, you can add `set -e` to a recipe line, like
```
@set -e; export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) ; cd $($@_SUBDIR) && $(MAKE) $(AM_MAKEFLAGS) install
```
then the line will fail if the `cd` fails, not just when the `$(MAKE)` fails.

fscheiner:
What I understood was, that even if a recipe line exits with !0, that does not mean, that `make` stops processing the remainder of the makefile or rule, e.g. for `make install` wouldn't it continue with installing other parts of the toolkit, when one part fails installing and just report that error? From the description in #118 I understood it that way.

And the patches should make sure that `make` stops execution as soon as an error happens IIUC.

matyasselmeci:
If you're running parallel make (`make -j`) then if one target fails, it's not going to kill the other targets. I'm not aware of a way around that. Other than `kill $$PPID` in the recipe line to kill the make process itself. That one sounds kinda nasty though :-)

fscheiner:
> If you're running parallel make (`make -j`) then if one target fails, it's not going to kill the other targets. I'm not aware of a way around that.

Ok, that could explain why it doesn't stop right away. #118 explicitly mentions the use of `-j 3` (both for `make` and `make install` - BTW, is the latter recommended?).

> Other than `kill $$PPID` in the recipe line to kill the make process itself. That one sounds kinda nasty though :-)

So @msalle's patches won't change that behaviour for parallel make runs?

matyasselmeci:
I wouldn't recommend parallel make for make install. make install is I/O bound anyway (since it's mostly file system operations), so adding `-j3` wouldn't give much of a speedup.

And yes, @msalle's patches won't help that case.  I think keeping the `exit 1`s in the `if` blocks is still useful though.

msalle:
Hi,
you're probably right about the non-parallel not needing it. The exit 1 does help for a make -j ... though

matyasselmeci:
Well it works for getting a nonzero error code, but it doesn't interrupt the make.  I ran some tests on my laptop (Fedora 31, make 4.2.1).

Here's my Makefile (`exittest.mk`):
```makefile
.PHONY: all error ok
all: error ok

error:
	exit 1

ok:
	sleep 5
	echo ok
```
and here's what happens if I run it:
```
$ make -j2 -f exittest.mk all; echo exit code $?
exit 1
sleep 5
make: *** [exittest.mk:6: error] Error 1
make: *** Waiting for unfinished jobs....
echo ok
ok
exit code 2
```

msalle:
Sure, it won't kill the running parallel jobs (see also https://www.gnu.org/software/make/manual/html_node/Parallel.html), but I think it is useful if at least the outer make process exit with a non-zero exit value if one of its child parallel jobs have failed?
Either way is fine with me, because as you said, make -j... install is generally not ideal, so feel free to close this PR and make a new one if you prefer.

matyasselmeci:
> I think it is useful if at least the outer make process exit with a non-zero exit value if one of its child parallel jobs have failed?
It already does: the exit status is 2.

msalle:
Hmm, you have a point there. I'm not sure what originally went wrong, but I agree it's probably indeed not necessary. Perhaps I got confused with the `if` which indeed would eat up the exit value. The other targets probably exit automatically when its make fails.
I'm afraid I don't have time reasonably soon to fix this and test this properly. If either of you have more time, please go ahead and fix in a better way.

PR #120, opened 2020-02-14
Make makefiles exit sooner on errors

matyasselmeci:
This is based on #119 but only keeps the `|| exit 1` in for loops and uses `&&` in a few places instead.

The changes are also applied to all the Makefile.am files that have those constructions (mostly `install-data-local` targets).


fscheiner:
Travis-CI check has passed and I also didn't see any problems during a local test run, so let's pull this one and close the other PR.

PR #122, opened 2020-03-11
Fedora test deps and cleanup

ellert:
Fedora changed the structure of their perl RPM and made it more modular. Some perl modules that used to be installed together with the perl interpreter are now separate modules. If these are used by the tests they must now be listed as build requires.

This PR also removes some unused test files.


ellert:
You are correct.
I have added perl-interpreter to all packages that use it either during build or testing. I also added all perl modules used and not only those not dragged in by other build dependencies.

PR #132, opened 2020-08-14
gridftp: Use example message from RFC 2428 for response to EPSV

fscheiner:
Title says it all.

Not a big change, but from now on the Globus GridFTP server should use the exact example message for a response to `EPSV` from [RFC 2428](https://tools.ietf.org/html/rfc2428), although the difference to a response to `PASV` could be already seen from the differences in the actual data part of the response (e.g. `(1,2,3,4,195,80)` (in response to `PASV`) and `(|||50000|)` (in response to `EPSV`)).

The difference to RFC 2428 was noticed during testing UberFTP 2.9-Beta against a Globus GridFTP server 13.20. UberFTP 2.9-Beta uses `EPSV` and `EPRT` by default (both commands are compatible with IPv4 and IPv6 address families) instead of `PASV` and `PORT` .

@msalle, @ellert: Sorry, but I'm again unsure about what to change in addition (major, minor version of "globus-gridftp-server-control" and "globus-gridftp-server", packaging information, library version) as this is not a functionality change and I don't expect clients to behave differently due to the changed message. Could you please help me out? :-)

### DIFF
--- a/gridftp/server-lib/src/configure.ac
+++ b/gridftp/server-lib/src/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gridftp_server_control],[9.0],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_gridftp_server_control],[9.1],[https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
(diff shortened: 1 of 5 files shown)


fscheiner:
@ellert: Thanks for making the needed changes to versions and packaging metadata.

PR #134, opened 2020-08-18
Add force option if the users do not exist in the machine (needed for e.g. EUDAT B2STAGE)

RubenGarcia:
When using gtc and EUDAT B2STAGE, the users may not be connected to the unix users in the machine (e.g. if authorization uses OpenID tokens).
I added an option to allow modification of the gridmap file even if the unix user does not exist (-f -- force)

### DIFF
--- a/gsi/gss_assist/source/programs/grid-mapfile-add-entry.8
+++ b/gsi/gss_assist/source/programs/grid-mapfile-add-entry.8
@@ -2,12 +2,12 @@
 .\"     Title: grid-mapfile-add-entry
 .\"    Author: [see the "AUTHOR" section]
 .\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
-.\"      Date: 06/12/2020
+.\"      Date: 08/21/2020
 .\"    Manual: Grid Community Toolkit Manual
 .\"    Source: Grid Community Toolkit 6
 .\"  Language: English
 .\"
-.TH "GRID\-MAPFILE\-ADD\-" "8" "06/12/2020" "Grid Community Toolkit 6" "Grid Community Toolkit Manual"
+.TH "GRID\-MAPFILE\-ADD\-" "8" "08/21/2020" "Grid Community Toolkit 6" "Grid Community Toolkit Manual"
 .\" -----------------------------------------------------------------
 .\" * Define some portability stuff
 .\" -----------------------------------------------------------------
@@ -87,7 +87,7 @@ Modify the gridmap file named by
 instead of the default\&.
 .RE
 .PP
-*\-force
+\fB\-force\fR
 .RS 4
 Make modifications even if user does not exist (needed for B2STAGE)\&.
 .RE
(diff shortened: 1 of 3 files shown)


fscheiner:
@matyasselmeci 
Strange, the Travis-CI check didn't work or didn't finish since Aug 18th, 3:00 PM GMT+2. I also don't see it on [Travis-CI](https://travis-ci.org/github/gridcf/gct/builds). Can this check be retriggered somehow?

gsi/gss_assist/source/programs/grid-mapfile-add-entry.8
```diff
@@ -87,6 +87,11 @@ Modify the gridmap file named by
 instead of the default\&.
 .RE
 .PP
+*\-force
```
fscheiner, line 90: I think this should read `\fB\-force\fR` when you compare it to the markup of `-dm -dryrun` further above. But maybe this is an artifact of the used "DocBook XSL Stylesheets v1.79.1" instead of what was used for the original file.

**UPDATE:** This is due to a missing `*` in https://github.com/gridcf/gct/pull/134/files#diff-970a287b14b8f63d7c4d1b1660881053. (see comment there!).

gsi/gss_assist/source/programs/grid-mapfile-add-entry.txt
```diff
@@ -65,6 +65,9 @@ The full set of command-line options to *grid-mapfile-add-entry* are:
 *-mapfile 'MAPFILE', -f 'MAPFILE'*::
     Modify the gridmap file named by 'MAPFILE' instead of the default.
 
+*-force::
```
fscheiner, line 68: Now I see the problem: A `*` is missing after `*-force`, then the manpage output should be created correctly.

fscheiner (changes requested):
@RubenGarcia 
Could you please make the change to https://github.com/gridcf/gct/pull/134/files#diff-970a287b14b8f63d7c4d1b1660881053, recreate the corresponding manpage and force-push your changes, so the PR gets updated?

I'd do it myself, but the GUI seems to not allow changes to your branch from me. Maybe because you create the PR from your master branch and not a "feature" branch.

fscheiner:
@RubenGarcia 
It could also be due to that you unchecked the "Allow edits by maintainers" checkbox when creating this PR, that I am unable to update your changes. I think allowing such changes in future PRs could speed up their inclusion, so better keep this checkbox always checked.

RubenGarcia:
I can't find how to change the option, so I'll perform your changes.

gsi/gss_assist/source/programs/grid-mapfile-add-entry.8
RubenGarcia: I added your changes and now the tests finish successfully.

gsi/gss_assist/source/programs/grid-mapfile-add-entry.txt
RubenGarcia: Changes applied.

fscheiner:
> I can't find how to change the option, so I'll perform your changes.

Indeed, I think this can only be activated when creating the PR.

fscheiner (approved):
Looks good to me.

Issue #112, opened 2019-12-16
globus-url-copy ignores GLOBUS_HOSTNAME; request to document GLOBUS_FTP_CLIENT_DATA_IP


RubenGarcia:
On a machine behind NAT, GLOBUS_HOSTNAME can normally be used to indicate the external IP to connect to. But globus-url-copy is sending the wrong IP to the server:

 /usr/local/globus-6/bin/globus-url-copy -vb -fast ftp://sunn-dtn.es.net:2811/data1/10G.dat file:///tmp/test.out
Source: ftp://sunn-dtn.es.net:2811/data1/
Dest:   file:///tmp/
  10G.dat  ->  test.out

            0 bytes         0.00 MB/sec avg         0.00 MB/sec inst
error: globus_ftp_client: the server responded with an error
500 500-Command failed. : globus_gridftp_server_file.c:globus_l_gfs_file_server_write_cb:3163:
500-callback failed.
500-globus_xio_tcp_driver.c:globus_l_xio_tcp_system_connect_cb:2022:
500-Unable to connect to 192.168.128.90:35309
500-globus_xio_system_select.c:globus_l_xio_system_handle_write:1108:
500-System error in connect: Network is unreachable
500-globus_xio: A system call failed: Network is unreachable
500 End.

export GLOBUS_HOSTNAME=138.246.234.17
/usr/local/globus-6/bin/globus-url-copy -vb -fast ftp://sunn-dtn.es.net:2811/data1/10G.dat file:///tmp/test.out
Source: ftp://sunn-dtn.es.net:2811/data1/
Dest:   file:///tmp/
  10G.dat  ->  test.out

            0 bytes         0.00 MB/sec avg         0.00 MB/sec inst
error: globus_ftp_client: the server responded with an error
500 500-Command failed. : globus_gridftp_server_file.c:globus_l_gfs_file_server_write_cb:3163:
500-callback failed.
500-globus_xio_tcp_driver.c:globus_l_xio_tcp_system_connect_cb:2022:
500-Unable to connect to 192.168.128.90:53139
500-globus_xio_system_select.c:globus_l_xio_system_handle_write:1108:
500-System error in connect: Network is unreachable
500-globus_xio: A system call failed: Network is unreachable
500 End.

/usr/local/globus-6/bin/globus-hostname 
138.246.234.17

RubenGarcia:
Also, GLOBUS_DOMAIN_NAME, used in common/source/library/globus_libc.c, is undocumented.

[...]

PR #135, opened 2020-08-21
Document environment variables in globus-url-copy manpage

RubenGarcia:
This solves
https://github.com/gridcf/gct/issues/112

### DIFF
--- a/gass/copy/source/globus-url-copy.txt
+++ b/gass/copy/source/globus-url-copy.txt
@@ -322,6 +322,30 @@ to a server that does not support data channel encryption (e.g. dCache)!
     transfer.  The default algorithm used for this checksum is MD5, but 
     other algorithms can be specified with the *-checksum-alg* parameter.
 
+[[globus-url-copy-ENVIRONMENT]]
+ENVIRONMENT
+-----------
+The following environment variables are recognized:
+
+*GLOBUS_HOSTNAME*::
+    Overrides client hostname communicated to the server
+
+*GLOBUS_DOMAIN_NAME*::
+    Overrides client domainname communicated to the server
+
+*GLOBUS_FTP_CLIENT_DATA_IP*::
+    Overwrites client IP address communicated to the server for the data
+    connection. Used when the receiver is behind a NAT and should be configured
+    to the external address of its gateway.
+
+*GLOBUS_FTP_CLIENT_SOURCE_PASV*::
+    Configures sender side (source) instead of receiver side (destination) to
+    open a data port and establishes the data connection from the receiver to
+    the sender. By default the data connection is established from sender to
+    receiver, but if the receiver is behind a NAT this doesn't work, because
+    the receiver will communicate its own address instead of the external
+    address of its gateway to the sender and its own address can't be reached
+    directly from the sender.
 
 [[globus-url-copy-AUTHOR]]
 AUTHOR
--- a/gass/copy/source/globus-url-copy.1
+++ b/gass/copy/source/globus-url-copy.1
@@ -2,12 +2,12 @@
 .\"     Title: globus-url-copy
 .\"    Author: [see the "AUTHOR" section]
 .\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
-.\"      Date: 07/25/2019
+.\"      Date: 09/08/2020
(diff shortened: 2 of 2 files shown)


gass/copy/source/globus-url-copy.txt
```diff
@@ -322,6 +322,20 @@ to a server that does not support data channel encryption (e.g. dCache)!
...
+
+*GLOBUS_HOSTNAME*::
+    Overrides hostname 
+
+*GLOBUS_DOMAIN_NAME*::
+    Overrides domainname
+
+*GLOBUS_FTP_CLIENT_DATA_IP*::
+    When using FTP transfer, the external IP of the machine (required when NAT is enabled)
+
```
fscheiner, line 338: I'm currently facing issues with the preview rendering of this source code comment in GitHub's GUI, so I don't know, how this will show. Hopefully it comes out correctly.

****

Looks good to me.

But I have a few questions before continuing:

Rereading https://github.com/gridcf/gct/issues/112#issue-538437049 I think using `GLOBUS_FTP_CLIENT_DATA_IP` is only needed when:

(1) you're doing a two-party transfer and the client is behind a NAT and you are using `-fast` which enables "MODE E", hence requiring that the sender/server connects to the receiver/client (i.e. like active FTP) and this can't work if the client is behind a NAT and provides its own address instead of its gateway's address. Without `-fast` (i.e. with "MODE S") you would need to use `GLOBUS_FTP_CLIENT_SOURCE_PASV` with an arbitrary value to make the client connect to the server instead, although the transfer is actually going from the server to the client.

...or:

(2) you're doing a 3rd party transfer and the receiving side is behind a NAT. Not sure if `GLOBUS_FTP_CLIENT_SOURCE_PASV` works in this case.

Please correct me, if your testing proves my assumptions above wrong.

To make that clear to a user, I think we should add which side these environment vars affect. I mean for `GLOBUS_FTP_CLIENT_DATA_IP` it seems to be clear, that this configures the IP address that the client communicates to the server for a data connection, but as `GLOBUS_FTP_CLIENT_SOURCE_PASV` is usually used to configure the sender side (source) to open a port and the receiver side (destination) to connect to this port - when the client is the receiver and behind a NAT - it might not always be correct to assume that an environment var starting with `GLOBUS_FTP_CLIENT` only affects the client side.

So what do you think about this changed text (I also included `GLOBUS_FTP_CLIENT_SOURCE_PASV` as it is related to the NAT case):

```
[[globus-url-copy-ENVIRONMENT]]
ENVIRONMENT
-----------
The following environment variables are recognized:

*GLOBUS_HOSTNAME*::
    Overrides client hostname communicated to the server

*GLOBUS_DOMAIN_NAME*::
    Overrides client domainname communicated to the server

*GLOBUS_FTP_CLIENT_DATA_IP*::
    Overwrites client IP address communicated to the server for the data connection. Used when the receiver is behind a NAT and should be configured to the external address of its gateway.

*GLOBUS_FTP_CLIENT_SOURCE_PASV*::
    Configures sender side (source) instead of receiver side (destination) to open a data port and establishes the data connection from the receiver to the sender. By default the data connections is established from sender to receiver, but if the receiver is behind a NAT this doesn't work, because the receiver will communicate its own address instead of the external address of its gateway to the sender and its own address can't be reached directly from the sender.
```

...? Of course there can be mutliple levels of gateways, but I hope it is clear what I mean with "external address of its gateway"?

I'd change that myself, but I again can't push commits to your repo, so you have to make the changes.

msalle: Just to add my support for the suggested further changes by @fscheiner

fscheiner: @RubenGarcia: And could you please this time **rebase** your feature branch onto the state of our master branch instead of **merging** in our recent changes? This because we can then avoid unnecessary commits in the GCT's commit history.

RubenGarcia: I just saw this. Looks good to me. If we find more corner cases, we can update the documentation later.
I'll update the code.

RubenGarcia: I added your changes and mixed everything into one commit. I hope all is ok now.

msalle (approved):
Looks good to me.

PR #136, opened 2020-08-25
Fix output payload check in gss_init_sec_context() after SSL handshake.

JaimeFrey:
When globus_i_gsi_gss_handshake() indicates that the SSL handshake has
completed successfully, the check for a final output payload to send is
faulty. It checks output_token before the data has been copied into that
buffer. Instead, it needs to look at context->gss_wbio.

This bug puts the client and server out of sync, dooming the connection.

### DIFF
--- a/gsi/gssapi/source/library/init_sec_context.c
+++ b/gsi/gssapi/source/library/init_sec_context.c
@@ -382,10 +382,13 @@ GSS_CALLCONV gss_init_sec_context(
          * have any more data to send, we can send the flag
          * now. i.e. fall through without break,
          * Otherwise, we will wait for the null byte
-         * to get back in sync which we will ignore
+         * to get back in sync which we will ignore.
+         * We can't check output_token->length, since we haven't
+         * extracted the output payload yet, so look directly
+         * in context->gss_wbio.
          */
 
-        if (output_token->length != 0)
+        if (BIO_pending(context->gss_wbio) != 0)
         {
             context->gss_state=GSS_CON_ST_FLAGS;
             break;


fscheiner:
@JaimeFrey: Ok, I restarted the Travis CI check. From what I saw in the logs, the original run exceeded the job time limit of 50 mins because some downloads during the test happened very slowly.

****

This looks like a good catch! :-) How did you detect that problem and do you have logs of what is going wrong without your changes and what differs with your changes applied?

JaimeFrey:
We noticed a test of GSI authentication within HTCondor failing on newer platforms (e.g. ubuntu18) but succeeding on older ones (e.g. rhel7). Comparing the network activity and calls within the gssapi library allowed me to spot what was going wrong. I rebuilt the GSI libraries with BUILD_DEBUG and then added some additional logging calls.

On rhel7, globus_i_gsi_gss_handshake() (and ultimately BIO_do_handshake()) indicates the handshake is complete on its third call on the client side, without no output data to send. On Ubuntu 18, it returns complete on the second call, with output data to send. Due to the bug, that data isn't sent.

I can attach logs if you're interested in the details.

fscheiner:
> We noticed a test of GSI authentication within HTCondor failing on newer platforms (e.g. ubuntu18) but succeeding on older ones (e.g. rhel7). Comparing the network activity and calls within the gssapi library allowed me to spot what was going wrong. I rebuilt the GSI libraries with BUILD_DEBUG and then added some additional logging calls.

Again, good catch! :+1: 

> On rhel7, globus_i_gsi_gss_handshake() (and ultimately BIO_do_handshake()) indicates the handshake is complete on its third call on the client side, without no output data to send. On Ubuntu 18, it returns complete on the second call, with output data to send. Due to the bug, that data isn't sent.

Just to be sure: this change doesn't break things on rhel7, right?

> I can attach logs if you're interested in the details.

If you already have them at hand or if it's not too much work to create them, these details could be interesting as background to this PR. But not a must, as I now think your explanation is already detailed enough.

JaimeFrey:
On rhel7, globus_i_gsi_gss_handshake() (and ultimately BIO_do_handshake()) indicates the handshake is complete on its third call on the client side, without no output data to send. On Ubuntu 18, it returns complete on the second call, with output data to send. Due to the bug, that data isn't sent.

Just to be sure: this change doesn't break things on rhel7, right?

Correct. On rhel7, the output buffer (context->gss_wbio) will be empty when the handshake is complete, resulting in the same behavior as before.


I can attach logs if you're interested in the details.

If you already have them at hand or if it's not too much work to create them, these details could be interesting as background to this PR. But not a must, as I now think your explanation is already detailed enough.

I will gather up the logs I have and attach them.

JaimeFrey:
It appears my description of the bug was a little wrong. The final data packet from the client for the SSL handshake is sent to the server (by globus_gss_assist_init_sec_context()). But gss_init_sec_context() indicates the full GSI authentication is complete, so globus_gss_assist_init_sec_context() doesn't call it again for the post-SSL-handshake data.

JaimeFrey:
I'm attaching the logs I have from my testing. This is from the condor_ping client tool, authenticating via GSI to a condor_schedd daemon. The logs include the gss_assist and gssapi debugging statements, as well as logging from our code. We provide our own network I/O functions (condor_read() and condor_write()), which are logging what they send/receive.

[rhel7-ok.log](https://github.com/gridcf/gct/files/5138406/rhel7-ok.log)
[ubuntu18-buggy.log](https://github.com/gridcf/gct/files/5138408/ubuntu18-buggy.log)
[ubuntu18-ok.log](https://github.com/gridcf/gct/files/5138409/ubuntu18-ok.log)

fscheiner:
Looks good to me and if @msalle, @ellert and @matyasselmeci don't object, I'd merge this. 

BTW the respective tests also succeed for the gsi/gssapi component according to https://api.travis-ci.org/v3/job/721008896/log.txt (search for e.g. "Testsuite summary for globus_gssapi_gsi"), but they also succeed for a test run under Ubuntu 18.04 according to my testing:

```
johndoe@ubuntu-18:~/git-projects/gridcf/gct/travis-ci$ sudo ./build_and_test.sh ubuntu ssh
[...]
make  check-TESTS
make[6]: Entering directory '/home/johndoe/git-projects/gridcf/gct/gsi/gssapi/source/test'
make[7]: Entering directory '/home/johndoe/git-projects/gridcf/gct/gsi/gssapi/source/test'
PASS: inquire-names-for-mech-test
PASS: compare-name-test-hybrid.pl
PASS: release-name-test
PASS: nonterminated-export-cred-test
PASS: gssapi-acquire-test
PASS: error-test
PASS: indicate-mechs-test
PASS: compare-name-test-rfc2818.pl
PASS: unwrap-null-test
SKIP: gssapi-import-context-test
PASS: mic-test
PASS: compare-name-test-gt2.pl
PASS: gssapi-inquire-sec-ctx-by-oid-test
PASS: tls-cipher-test
PASS: gssapi_expimp_cred_test
PASS: gssapi-anonymous-test
PASS: alpn-test
SKIP: gssapi-expimp-test
PASS: sni-test
PASS: wrap-test
PASS: duplicate-name-test
PASS: gssapi-import-name
PASS: gssapi-thread-test-wrapper
PASS: tls-version-test
PASS: mech-compatibility-test.pl
PASS: gssapi-delegation-test
PASS: gssapi-limited-delegation-test
PASS: import-cred-test.pl
PASS: gssapi-delegation-compat-test
============================================================================
Testsuite summary for globus_gssapi_gsi 14.13
============================================================================
# TOTAL: 29
# PASS:  27
# SKIP:  2
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
[...]
```

So the actual issue is not covered by the activated tests for this component.

msalle (approved):
@msalle

Hi,
I first thought you might need to use BIO_wpending() instead of BIO_pending() but I think it's correct after all:
- BIO_pending() and BIO_wpending() are macros calling BIO_ctrl() using cmd equal to either BIO_CTRL_PENDING or BIO_CTRL_WPENDING
- running through gdb it seems the context->wbio is a "memory buffer" BIO, as defined in bss_mem.c
- looking then at mem_ctrl() it looks like bio_pending() is indeed the right one since for BIO_CTRL_WPENDING it would always return 0: bss_mem.c#L326_L331.

Unfortunately the man pages are not very informative about the difference between the two )-:

PR #142, opened 2020-12-11
Temporarily disable deployment to GitHub Releases

matyasselmeci:


fscheiner (approved):
OK, let's try this. The old version fails with or without an existing release.

PR #143, opened 2020-12-17
Keep admin script in sync with init script

ellert:
Even though there is a comment in the admin script that says:
  # Keep in sync with init script
the admin script is not in sync with the init script.
This commit corrects this.

The admin script ignored the GLOBUS_SEG_PIDFMT setting in the
sysconfig file that the init script uses. With the default value they
still did the same, and there is usually no need to change the
default, so it usually worked anyway.

### DIFF
--- a/gram/jobmanager/scheduler_event_generator/source/configure.ac
+++ b/gram/jobmanager/scheduler_event_generator/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_scheduler_event_generator],[6.3],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_scheduler_event_generator],[6.4],[https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
 AC_SUBST([MAJOR_VERSION], [${PACKAGE_VERSION%%.*}])
 AC_SUBST([MINOR_VERSION], [${PACKAGE_VERSION##*.}])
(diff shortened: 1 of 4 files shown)


matyasselmeci:
Doesn't this now ignore GLOBUS_SEG_PIDDIR? Is that set anywhere?

ellert:
> Doesn't this now ignore GLOBUS_SEG_PIDDIR? Is that set anywhere?

This is the only place GLOBUS_SEG_PIDDIR is mentioned in the entire source tree.
It is not set in the default version of the sysconfig file, which does set GLOBUS_SEG_PIDFMT.
https://github.com/gridcf/gct/blob/6259fe338390a1183401524772b2acaf4baf06e7/gram/jobmanager/scheduler_event_generator/source/config/globus-scheduler-event-generator.in#L1
It is possible to set GLOBUS_SEG_PIDDIR by adding it to the sysconfig file. However, if that is done this setting is ignored by the init script which only supports GLOBUS_SEG_PIDFMT.
https://github.com/gridcf/gct/blob/6259fe338390a1183401524772b2acaf4baf06e7/gram/jobmanager/scheduler_event_generator/source/init/globus-scheduler-event-generator-lsb.in#L73
Similarly, changing the default values of GLOBUS_SEG_PIDFMT specified in the sysconfig file will cause a change in where the init script says that the pid file is, but will be ignored by the admin script.
With this proposed change, both the init script and the admin script will recognise the GLOBUS_SEG_PIDFMT variable, which is the one that is mentioned in the default sysconfig file. (And both will ignore GLOBUS_SEG_PIDDIR.)

matyasselmeci (approved):

matyasselmeci:
OK, looks good.

PR #147, opened 2021-02-25
exit with error if voms-proxy-init fails

### LINKED ISSUE #68: myproxy-logon still returns code 0 when voms-proxy-init called by it fails
Return value from voms_proxy_init() is checked for negative values instead of non-zero values.

jbasney:
fixes #68

### DIFF
--- a/myproxy/source/myproxy_get_delegation.c
+++ b/myproxy/source/myproxy_get_delegation.c
@@ -203,9 +203,10 @@ main(int argc, char *argv[])
 
     if (outputfile) {
         if (voms && (! has_voms_extension(outputfile)) ) {
-            if (voms_proxy_init() < 0) { /* should an error be fatal? */
-                fprintf(stderr, "Warning: Failed to add VOMS attributes.\n");
+            if (voms_proxy_init() != 0) {
+                fprintf(stderr, "Failed to add VOMS attributes.\n");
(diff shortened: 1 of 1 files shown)


fscheiner:
@jbasney
Ok, I also did some renaming (`rc`(as used by `waitpid()`) => `wstatus`) but kept `rc` as "returncode" for `voms_proxy_init()` but the logic stays the same.

fscheiner (approved):
Looks good to me.

fscheiner:
@matyasselmeci 
In the "checks" part in https://github.com/gridcf/gct/pull/147, there are two lines shown currently, one for "continuous-integration/travis-ci" and one for "Travis CI - Pull Request", where the latter seems to be the "new" one using `travis-ci-com` instead of `travis-ci.org`. The first check seems to not run.

Do we need to remove the first check from the (repository) configuration somewhere?

matyasselmeci:
@fscheiner, I followed the instructions here
https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154
and it seems to work now.

fscheiner:
@matyasselmeci 
Great! Thanks for solving this.

Issue #43, opened 2018-05-23
globus-gssapi-gsi and TLS 1.3


ellert:
The upcoming openssl 1.1.1 adds support for TLS 1.3. However globus-gssapi-gsi is incompatible with TLS 1.3, as can be seen from the results of "make check".

See globus/globus-toolkit#123 for details.

Openssl 1.1.1 is available in Debian experimental.

Any TLS experts available to investigate?

fscheiner:
The following is a list of files in our repo at 4361e1e307eeb9ee73c7523d8bea19aafa674a45 that seem to include OpenSSL headers:
```
gct$ grep -nr '#include <openssl/' * | cut -d ':' -f1 | sort -u
gass/copy/source/globus_gass_copy.c
gass/copy/source/globus_gass_copy.h
gass/copy/source/globus_url_copy.c
gram/jobmanager/source/globus_gram_job_manager_gsi.c
gridftp/hdfs/src/gridftp_hdfs_cksm.c
gridftp/hdfs/src/gridftp_hdfs.h
gridftp/server/src/globus_i_gfs_data.c
gridftp/server/src/modules/file/globus_gridftp_server_file.c
gsi/gridmap_eppn_callout/globus_gridmap_eppn.c
gsi/gridmap_verify_myproxy_callout/globus_gridmap_verify_myproxy.c
gsi/gssapi/source/configure.ac
gsi/gssapi/source/test/gssapi_inquire_sec_ctx_by_oid_test.c
gsi/gssapi/source/test/mech_compatibility_test.c
gsi/proxy/proxy_ssl/source/library/proxycertinfo.c
gsi/proxy/proxy_ssl/source/library/proxycertinfo.h
gsi/proxy/proxy_ssl/source/library/proxypolicy.c
gsi/proxy/proxy_ssl/source/library/proxypolicy.h
gsi/proxy/proxy_ssl/source/test/test_proxycertinfo.c
gsi/proxy/proxy_utils/source/programs/grid-cert-diagnostics.c
myproxy/source/certauth_extensions.c
myproxy/source/myproxy_common.h
myproxy/source/myproxy_ocsp_aia.c
myproxy/source/myproxy_ocsp_aia.h
myproxy/source/myproxy_ocsp.c
myproxy/source/myproxy_ocsp.h
myproxy/source/ssl_utils.h
myproxy/source/voms_utils.h
packaging/debian/gsi-openssh/debian/patches/hpn-14.13-isshd.v3.19.1.patch
[...]

PR #149, opened 2021-03-04
Fix name matching for Doxygen

fscheiner:
...by avoiding file names for the `\file` command in the files
changed with this commit. See https://www.doxygen.nl/manual/commands.html#cmdfile
for details.

Fixes #109.

PR #150, opened 2021-03-25
TLS v1.3

ellert:
Make GCT compatible with TLS v1.3.

With this change I was able to rebuild all other GCT packages without changes aigainst the updated globus-gssapi-gsi that enables TLS v1.3. That is all the unit tests that are run during the package build succeeded with TLS v1.3.


fscheiner:
Just to be sure: If we merge #146 and/or #149 before this one, do you need to rebase this one? If yes, I'd prefer to merge this one here first.

ellert:
> Just to be sure: If we merge #146 and/or #149 before this one, do you need to rebase this one? If yes, I'd prefer to merge this one here first.

There are no conflicts with #146. There is one common file with #149, but the changes are far apart, so either being merged before the other should not cause conflicts. No need for rebase.

fscheiner (approved):
As per https://github.com/gridcf/gct/pull/150#issue-600845616 @ellert has confirmed that this works, so let's merge this and close [our second oldest open issue](https://github.com/gridcf/gct/issues/43) and make the GCT fit for the future.

Maybe these changes also justify a new release, what do you think?

maarten-litmaath:
Hi guys,
this is great news: congratulations & thanks very much!

However, we will need to be careful with releasing the new code,
as it has not been tested against production services, particularly
those that have their own GSI stack, like dCache and XRootD.

I think we need to keep the maximum TLS version at 1.2 by default,
so that things will keep working as before, and then we can ask for
v1.3 to be tried by various teams, e.g. the FTS devs.

What do you think?

ellert:
It obviously needs to be tested against other implementations.
Note that TLS 1.3 requires openssl 1.1.
The default version of openssl in RHEL/CentOS 7 is openssl 1.0, so any installation running on that will be limited to TLS 1.2 anyway.
Whether we should set the default higher version to 1.2 can be decided later based on what testing we manage to do before the next release. This just means changing a line in gsi.conf - no changes needed to the code.

maarten-litmaath:
Excellent, thanks!

fscheiner:
Fixes #43.

PR #155, opened 2021-06-15
Compatibility with the dcache server implementation

ellert:
See https://github.com/dCache/dcache/issues/5939.

### DIFF
--- a/gsi/gssapi/source/library/init_sec_context.c
+++ b/gsi/gssapi/source/library/init_sec_context.c
@@ -202,11 +202,8 @@ GSS_CALLCONV gss_init_sec_context(
             *ret_flags = 0 ;
         }
     }
-    else
+    else if (input_token->length > 0)
     {
-        /* first time there is no input token, but after that
-         * there will always be one
-         */
         major_status = globus_i_gsi_gss_put_token(&local_minor_status,
                                                   context,
                                                   NULL,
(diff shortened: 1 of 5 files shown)


gsi/gssapi/source/library/init_sec_context.c
```diff
@@ -202,6 +202,13 @@ GSS_CALLCONV gss_init_sec_context(
             *ret_flags = 0 ;
         }
     }
+    else if (context->gss_state == GSS_CON_ST_TOKEN1 &&
+             input_token->length == 0) {
+        /*
+         * This is for compatibility with the dcache server implementation
+         */
+        context->gss_state = GSS_CON_ST_FLAGS;
+    }
```
fscheiner, line 211: @ellert:
IIUC this forwards the connection state (as defined in https://github.com/gridcf/gct/blob/f8ccd9058e9d54d0cc9430818b0a4a9bd65b9b1d/gsi/gssapi/source/library/globus_gsi_gss_constants.h#L94..L106) for the client by skipping the `GSS_CON_ST_TOKEN2` state if no or an empty input token follows the output token that follows the first input token. This enables the GCT against dCache case when TLS v1.3 is used but doesn't harm the GCT against GCT case for TLS v1.3, right?

Why not include the skipping below https://github.com/gridcf/gct/blob/5a239381ad4444a8e96c9d2154251586f55f48f9/gsi/gssapi/source/library/init_sec_context.c#L421 ...?

ellert: I have rewritten the code according to your suggestion.

fscheiner: Thanks.

fscheiner (approved):
@ellert:
Looks good to me, but to be honest I can't say much about the detailed functionality of the recent code changes, you're definitely the expert here.

But testing a locally compiled `globus-url-copy` - with your changes included - against the dCache instance at "prometheus.desy.de" works for both TLS v1.2 and v1.3 (file listing tested). 

The same holds true when testing against the GridFTP server version in openSUSE Leap 15.2.

The same holds true when testing against the GridFTP server version in CentOS 6.10 (when allowing usage of TLS v1.3 it correctly falls back to TLS v1.2 for the connection).

Therefore these changes solve https://github.com/dCache/dcache/issues/5939 for me and don't change functionality for the GCT GridFTP servers tested.

So let's merge this.

fscheiner:
Fixes https://github.com/dCache/dcache/issues/5939.

PR #156, opened 2021-06-22
Fix for segmentation fault with tilde:

ellert:
```
globus-url-copy -list gsiftp://localhost/~
gsiftp://localhost/~
Segmentation fault (core dumped)
```


msalle (approved):
Looks good to me, thanks!

PR #157, opened 2021-06-28
Add fixes for OpenSSL-3.0.0-beta1

msalle:
A few small changes make the GCT build against OpenSSL-3.0.0-beta1. I've tested some parts run-time.

- several function parameters are now `const`
- `CRYPTO_mem_ctrl()` (together with several others) is deprecated and by default no longer available, unless OpenSSL is configured using `enable-crypto-mdebug`


gsi/gssapi/source/test/gssapi_anonymous_test.c
```diff
@@ -60,8 +60,15 @@ int main()
     
     /* acquire the credential */
 
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
```
fscheiner, line 63: @msalle:
OpenSSL uses [`#  ifndef OPENSSL_NO_DEPRECATED_3_0`](https://github.com/openssl/openssl/blob/b8be229dab036b26de8830444bf2beb82e71f50e/include/openssl/crypto.h.in#L361), not sure what's the better way. But I definitely find the version check used in your PR much better comprehensible.

gsi/proxy/proxy_core/source/library/globus_gsi_proxy.c
```diff
@@ -136,7 +136,14 @@ globus_l_gsi_proxy_activate(void)
         CRYPTO_malloc_debug_init();
         CRYPTO_dbg_set_options(V_CRYPTO_MDEBUG_ALL);
 #endif
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
+/* From CRYPTO_mem_ctrl.3ossl man page: The memory-leak checking has been
+ * deprecated in OpenSSL 3.0 in favor of clang's memory and leak sanitizer.
+ * Alternatively, openssl needs to be configured using
+ *  ./Configure enable-crypto-mdebug
+ */
         CRYPTO_mem_ctrl(CRYPTO_MEM_CHECK_ON);
```
fscheiner, line 145: @msalle:
`CRYPTO_mem_ctrl(CRYPTO_MEM_CHECK_ENABLE);` is "missing" here, compared to the test case above. From https://github.com/openssl/openssl/blob/b8be229dab036b26de8830444bf2beb82e71f50e/include/openssl/crypto.h.in#L349..L352 I can't make up what the "correct" sequence is if a sequence is needed at all.

gsi/gssapi/source/test/gssapi_anonymous_test.c
fscheiner: The [corresponding documentation](https://www.openssl.org/docs/manmaster/man7/OPENSSL_NO_DEPRECATED.html) also doesn't make it clearer IMHO.

gsi/proxy/proxy_ssl/source/library/proxycertinfo.h
```diff
@@ -154,6 +154,9 @@ ASN1_METHOD * PROXYCERTINFO_asn1_meth();
 
 GLOBUS_DEPRECATED(
 PROXYCERTINFO * PROXYCERTINFO_dup(
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+    const
+#endif
     PROXYCERTINFO *                     cert_info));
```
fscheiner, line 160: @msalle:
Sorry, but why do we need a `const` here (and in the file below) with OpenSSL 3.0.0? I can't find the `PROXYCERTINFO_dup()` function in [OpenSSL's repo](https://github.com/openssl/openssl) nor did I find the `PROXYCERTINFO` type there.

gsi/proxy/proxy_ssl/source/library/proxypolicy.h
```diff
@@ -104,6 +104,9 @@ ASN1_METHOD * PROXYPOLICY_asn1_meth();
 #endif
 
 PROXYPOLICY * PROXYPOLICY_dup(
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+    const
+#endif
     PROXYPOLICY *                       policy);
```
fscheiner, line 110: @msalle:
There are additional occurances of non-const `PROXYPOLICY` pointers below. Why don't they get a `const` type qualifier added for OpenSSL 3.0.0?

fscheiner (approved):
Looks good to me, but I have a few questions below.

Let's merge that after #155.

gsi/gssapi/source/test/gssapi_anonymous_test.c
msalle: I think for us it's much better to stick to a direct check against the version. That has always worked and is therefore more reliable.
For background, see e.g. https://github.com/openssl/openssl/blob/master/include/openssl/crypto.h.in#L361-L363 and https://github.com/openssl/openssl/blob/master/include/openssl/macros.h#L159-L190 plus https://github.com/openssl/openssl/blob/master/doc/internal/man7/deprecation.pod

gsi/proxy/proxy_core/source/library/globus_gsi_proxy.c
msalle: I agree it's not clear and I think I'd prefer to just not touch that part since it's unrelated to OpenSSL3.0 fixing.

gsi/proxy/proxy_ssl/source/library/proxycertinfo.h
msalle: They are defined around https://github.com/gridcf/gct/blob/master/gsi/proxy/proxy_ssl/source/library/proxycertinfo.h#L136 via a bunch of macros in https://github.com/openssl/openssl/blob/master/include/openssl/asn1.h.in
So in this case for the name `PROXYCERTINFO`. The `*_dup()` function now wants a `const`, see https://github.com/openssl/openssl/blob/master/include/openssl/asn1.h.in#L289 and commit https://github.com/openssl/openssl/commit/9fdcc21fdc9d148f78d9cd5be34030f38cc45812

gsi/proxy/proxy_ssl/source/library/proxypolicy.h
msalle: I fixed the ones that gave errors, I think the rest is ok or not coming from OpenSSL?
We probably could make `PROXYPOLICY_print()` and `PROXYPOLICY_print_fp()` use a const, but they are defined in gridcf, not openssl.
Did you have a specific one in mind?

gsi/gssapi/source/test/gssapi_anonymous_test.c
fscheiner: I fully agree and it's easier to understand anyhow.

gsi/proxy/proxy_core/source/library/globus_gsi_proxy.c
fscheiner: Agreed.

gsi/proxy/proxy_ssl/source/library/proxycertinfo.h
fscheiner: Ok, thanks for the explanation.

gsi/proxy/proxy_ssl/source/library/proxypolicy.h
fscheiner: No, actually not. I tried to recognize a pattern for this change, but failed, so asked to be sure.

I assumed if `PROXYCERTINFO_dup()` "comes" from OpenSSL, than also the other `PROXYCERTINFO_[...]()` functions come from there and when the `PROXYCERTINFO * [...]` argument is `const` for `PROXYCERTINFO_dup()` with OpenSSL 3.0 it should also be `const` for the other functions where it made sense from their naming.

Now seeing that all but `PROXYCERTINFO_dup()` have an "obvious" implementation in [gsi/proxy/proxy_ssl/source/library/proxycertinfo.c](https://github.com/gridcf/gct/blob/5882044a0eac96b38b276f0ed4f565afc16fe494/gsi/proxy/proxy_ssl/source/library/proxycertinfo.c) makes it a little clearer.

fscheiner:
@msalle @matyasselmeci @ellert:
BTW, is it only me or do others also find that OpenSSL 3.0 reminds one of the unsecure SSL version 3?

maarten-litmaath:
Yeah, they should have gone straight to v4 instead...  :-)

fscheiner:
Indeed, there is a reason why Microsoft went straight to 10 from 8.1 - nobody wanted to see a Windows 9.5 or 9.8.

Anyone knows why OpenSSL developers didn't use 2.0? Is 2.0 an unlucky number (for them)?

msalle:
> Indeed, there is a reason why Microsoft went straight to 10 from 8.1 - nobody wanted to see a Windows 9.5 or 9.8.
> 
> Anyone knows why OpenSSL developers didn't use 2.0? Is 2.0 an unlucky number (for them)?

See [The Holy Hand Grenade of Antioch](https://www.openssl.org/blog/blog/2018/11/28/version/) (-;

fscheiner:
> See [The Holy Hand Grenade of Antioch](https://www.openssl.org/blog/blog/2018/11/28/version/) (-;

Thanks for that. [The next article there](https://www.openssl.org/blog/blog/2018/12/20/20years/) explains why it's OpenSSL and not OpenTLS.

fscheiner:
#155 is in, so I'm gonna merge this one now.

PR #154, opened 2021-06-10
Include GSI-OpenSSH 8.6p1 w/HPN from Fedora 34

fscheiner:
Should fix #67. Based on ~gsi-openssh-8.6p1-1.fc34~ gsi-openssh-8.6p1-2.fc34

****

This PR so far includes the fully patched source code of GSI-OpenSSH 8.6p1 w/HPN from Fedora 34.

This was tested to compile and work correctly (with [test-gss-kex-for-gsi-openssh.bash](https://gist.github.com/fscheiner/92ea125c72cd70283a712585206c1015)) on:

* openSUSE Leap 15.2:

```
johndoe@host:~> sudo ~/bin/test-gss-kex-for-gsi-openssh.bash host.domain.tld johndoe2
gsisshd: OpenSSH_8.6p1c-GSI GSI-hpn15v2, OpenSSL 1.1.1d  10 Sep 2019
gsissh: OpenSSH_8.6p1c-GSI GSI-hpn15v2, OpenSSL 1.1.1d  10 Sep 2019

Wait 3 seconds for startup of gsisshd ...

gss-gex-sha1- OK
gss-group1-sha1- OK
gss-group14-sha256- OK
gss-nistp256-sha256- OK
gss-curve25519-sha256- OK
gss-group16-sha512- OK
```

* Ubuntu 18.04.5:

```
johndoe@host:~$ sudo bin/test-gss-kex-for-gsi-openssh.bash host.domain.tld johndoe2
gsisshd: OpenSSH_8.6p1c-GSI GSI-hpn15v2, OpenSSL 1.1.1  11 Sep 2018
gsissh: OpenSSH_8.6p1c-GSI GSI-hpn15v2, OpenSSL 1.1.1  11 Sep 2018

Wait 3 seconds for startup of gsisshd ...

gss-gex-sha1- OK
gss-group1-sha1- OK
gss-group14-sha256- OK
gss-nistp256-sha256- OK
gss-curve25519-sha256- OK
gss-group16-sha512- OK
``` 

...so I expect it to work on other Linux distributions (SLES, Debian GNU/Linux, etc.) as well, of course incl. anything RHEL/Fedora based.

****

Configure with e.g.:

```
$ ./configure \
--sysconfdir=/usr/local/etc/gsissh \
--libexecdir=/usr/local/libexec/gsissh \
--datadir=/usr/local/share/gsissh \
--with-default-path=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin \
[...]

PR #158, opened 2021-08-08
Fix for gsi openssh builds on centos 7

fscheiner:
As per https://github.com/gridcf/gct/pull/154#issuecomment-894463701 and following comments, this PR should fix the build problems introduced with #154.

Again much thanks to @msalle for providing a patch that makes GSI-OpenSSH w/HPN 8.6p1 compile with OpenSSL version 1.0.x (as shipped with CentOS 7, which is used on Travis CI for our build tests).

### DIFF
--- a/gsi_openssh/source/ssh-rsa.c
+++ b/gsi_openssh/source/ssh-rsa.c
@@ -53,6 +53,16 @@ rsa_hash_alg_ident(int hash_alg)
 	return NULL;
 }
 
+#if (OPENSSL_VERSION_NUMBER < 0x10100000L)
+static int RSA_bits(const RSA *r);
+
+static int
+RSA_bits(const RSA *r)
+{
+    return BN_num_bits(r->n);
+}
+#endif
+
 /*
  * Returns the hash algorithm ID for a given algorithm identifier as used
  * inside the signature blob,
--- a/travis-ci/build_and_test.sh
+++ b/travis-ci/build_and_test.sh
@@ -10,7 +10,6 @@ cd "$root"
 
 args=(--prefix="$root" --enable-silent-rules)
 if [[ $COMPONENTS != *ssh* ]]; then
-    rm -f prep-gsissh
     args+=(--disable-gsi-openssh)
 fi
 if [[ $COMPONENTS == *gram5* ]]; then
--- a/travis-ci/run_task_inside_docker.sh
+++ b/travis-ci/run_task_inside_docker.sh
@@ -30,7 +30,8 @@ yum clean all
 packages=(gcc gcc-c++ make autoconf automake libtool \
           libtool-ltdl-devel openssl openssl-devel git \
           'perl(Test)' 'perl(Test::More)' 'perl(File::Spec)' \
-          'perl(URI)' file sudo bison patch curl)
+          'perl(URI)' file sudo bison patch curl \
+          pam pam-devel libedit libedit-devel)
 
 if [[ $TASK == tests ]]; then
     set +e
@@ -56,6 +57,8 @@ elif [[ $TASK == *rpms ]]; then
     packages+=(redhat-lsb-core)
     # for myproxy-oauth
     packages+=(m2crypto mod_ssl mod_wsgi pyOpenSSL python-crypto)
+    # for gsi-openssh
+    packages+=(pam libedit libedit-devel)
 fi
 
 


msalle (commented):
Out of curiosity, what's your reason to remove PAM and libedit?
I did manage to compile with the flags enabled.

fscheiner:
> Out of curiosity, what's your reason to remove PAM and libedit?

We don't install the corresponding devel packages during the build tests. I wanted to leave this stuff untouched, so removed these dependencies instead. And it looks like PAM is also not needed for GSI authentication. libedit would only be needed for command line editing when using `sftp`. I didn't consider that too important because users can still change the configure defaults for GSI-OpenSSH when they compile it locally.

> I did manage to compile with the flags enabled.

Most likely because you have the corresponding devel packages installed, like me during development.

fscheiner:
Ok, I'm merging this now to fix our build tests.

msalle:
> > Out of curiosity, what's your reason to remove PAM and libedit?
> 
> We don't install the corresponding devel packages during the build tests. I wanted to leave this stuff untouched, so removed these dependencies instead. And it looks like PAM is also not needed for GSI authentication. libedit would only be needed for command line editing when using `sftp`. I didn't consider that too important because users can still change the configure defaults for GSI-OpenSSH when they compile it locally.

true, but for gsiftp command-line editing seems a useful default to have?
Whether PAM is needed depends I think. If gsissh should fail over to e.g. password login, PAM can be very useful. Also for several other scenarios probably even in combination with GSI.
From a quick look, I'd say we'd need to add them to [run_task_inside_docker.sh](https://github.com/gridcf/gct/blob/master/travis-ci/run_task_inside_docker.sh), but I'm not sure whether to add to the generic `packages` in line 30 and further or to the `*rpms` in line 41 and further.
Obviously @matyasselmeci will know much better.

> 
> > I did manage to compile with the flags enabled.
> 
> Most likely because you have the corresponding devel packages installed, like me during development.

indeed I did. Actually, I installed them for this purpose (-;

fscheiner:
> true, but for gsiftp command-line editing seems a useful default to have?

Hey, who uses gsisftp if GridFTP is available? And [gtransfer](http://gtransfer.org/) has all the goodies (command completion, remote directory browsing, etc.) just with bash completion. ;-)

> Whether PAM is needed depends I think. If gsissh should fail over to e.g. password login, PAM can be very useful. Also for several other scenarios probably even in combination with GSI.

Doesn't this work without PAM as long as password authentication is still enabled in `sshd_config`?

I don't have a strong opinion on both, but as I would like to include the new GSI-OpenSSH in our next release and that should materialize in week 33 we need to get this done next week if we want to keep PAM and libedit.

> From a quick look, I'd say we'd need to add them to [run_task_inside_docker.sh](https://github.com/gridcf/gct/blob/master/travis-ci/run_task_inside_docker.sh), but I'm not sure whether to add to the generic `packages` in line 30 and further or to the `*rpms` in line 41 and further.

I'd go for the latter, so it's clear PAM and libedit are for GSI-OpenSSH.

****

If you (@msalle, @matyasselmeci, @ellert, @maarten-litmaath) don't mind our build tests being "broken" until then, I'll try to make all that happen next week. It would be better for our Travis CI credits anyhow if we do everything at the same time, not that we end up with too less credits for our release build (we have 4500 credits left and each "normal" build costs about 840 credits, a release build most likely more).

What do you think?

maarten-litmaath:
Hi all,
I think we do not need to include any new functionality,
just bugfixes for components used in production...

msalle:
> Hey, who uses gsisftp if GridFTP is available? And [gtransfer](http://gtransfer.org/) has all the goodies (command completion, remote directory browsing, etc.) just with bash completion. ;-)

actually, I have used gsiftp quite often to copy files out of a system which had no gridftp available, only gsissh.

> > Whether PAM is needed depends I think. If gsissh should fail over to e.g. password login, PAM can be very useful. Also for several other scenarios probably even in combination with GSI.
> 
> Doesn't this work without PAM as long as password authentication is still enabled in `sshd_config`?

it does, but password auth without pam is generally rather limited and unpractical. And there are of course all kinds of different useful PAM modules.

> I don't have a strong opinion on both, but as I would like to include the new GSI-OpenSSH in our next release and that should materialize in week 33 we need to get this done next week if we want to keep PAM and libedit.

I understand your reasoning but I personally think it would be a bit of a pity not to include these settings, since it is only a matter of installing the build-time requirements. Also note (to reply to @maarten-litmaath 's comment)  that this is not about adding but removing current functionality.

> > From a quick look, I'd say we'd need to add them to [run_task_inside_docker.sh](https://github.com/gridcf/gct/blob/master/travis-ci/run_task_inside_docker.sh), but I'm not sure whether to add to the generic `packages` in line 30 and further or to the `*rpms` in line 41 and further.
> 
> I'd go for the latter, so it's clear PAM and libedit are for GSI-OpenSSH.

I agree that would be logical, but it seems that one might be only for the RPMs? I haven't figured out how it actually worked. 

> If you (@msalle, @matyasselmeci, @ellert, @maarten-litmaath) don't mind our build tests being "broken" until then, I'll try to make all that happen next week. It would be better for our Travis CI credits anyhow if we do everything at the same time, not that we end up with too less credits for our release build (we have 4500 credits left and each "normal" build costs about 840 credits, a release build most likely more).
> 
> What do you think?

As above, I'd prefer a test to fail than to have not ideal build configuration defaults which furthermore is a change in what we have now.

fscheiner:
PAM and libedit dependencies are in again.

fscheiner:
> > From a quick look, I'd say we'd need to add them to [run_task_inside_docker.sh](https://github.com/gridcf/gct/blob/master/travis-ci/run_task_inside_docker.sh), but I'm not sure whether to add to the generic `packages` in line 30 and further or to the `*rpms` in line 41 and further.
> 
> I'd go for the latter, so it's clear PAM and libedit are for GSI-OpenSSH.

Ok, they need to be added to the generic list of packages (in line 30), as gsi_openssh gets configured in all cases and installing PAM and libedit dependencies only when component `*ssh*` is used - as it is now with 840e9d1 - is not enough then. I'll make this change and retest.

fscheiner:
This PR should be complete now. The `configure` steps run through, the `make` runs should continue to work (fingers crossed), as I didn't change anything there. I cancelled the builds on Travis CI afterwards to save credits. Will merge that now.

Issue #160, opened 2021-08-19
myproxy FTBFS on IPV6-only build servers


ellert:
Forwarded from:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=992436

Anyone has any idea how to fix this?

fscheiner:
CCing @jbasney 

Strange. Comparing the [Debian build log](https://buildd.debian.org/status/fetch.php?pkg=myproxy&arch=amd64&ver=6.2.8-1&stamp=1629290865&raw=0) to [my log on the OBS](https://build.opensuse.org/public/build/home:frank_scheiner:gct/openSUSE_Leap_15.2/x86_64/myproxy/_log) (though for MyProxy v6.2.**6**, but the changes in #146 and #147 don't look like they changed something relevant), it looks to me, like there's just no service listening on the IPv6 localhost address on the Debian build host, because the server seems to listen on the IPv4 localhost address:

```
[...]
Attempting to connect to ::1:45721 
Attempting to connect to 127.0.0.1:45721 
Successfully connected to localhost:45721
[...]
```

Maybe `myproxy-server` (commandline in https://github.com/gridcf/gct/blob/master/myproxy/source/myproxy-test#L269) doesn't listen on IPv6 addresses by default, when the env var `MYPROXY_SERVER` is configured to `localhost` only (done in https://github.com/gridcf/gct/blob/master/myproxy/source/myproxy-test#L243). On Ubuntu 20.04 for example the IPv6 localhost address has the following aliases:

*  `ip6-localhost` 
*  `ip6-loopback`

Not sure if (1) we just would need to add one of these to `MYPROXY_SERVER` (@jbasney: Can this be done and what separator is needed here?) or (2) we need to add additional `-l [...]` arguments to the server command in https://github.com/gridcf/gct/blob/master/myproxy/source/myproxy-test#L269) to solve this.

fscheiner:
Additonal info:

"Bootstrapping MyProxy server root of trust." comes from https://github.com/gridcf/gct/blob/5882044a0eac96b38b276f0ed4f565afc16fe494/myproxy/source/myproxy.c#L497
"Attempting to connect to [...]"  comes from https://github.com/gridcf/gct/blob/5882044a0eac96b38b276f0ed4f565afc16fe494/myproxy/source/myproxy.c#L322

jbasney:
`MYPROXY_SERVER` can be a comma-separated list of hostnames. That tells clients where to connect.
[...]

PR #161, opened 2021-08-20
Use -l (--listen) flag when starting myproxy-server in test scripts

ellert:
Fix for #160. As proposed in https://github.com/gridcf/gct/issues/160#issuecomment-901860742.


msalle (approved):
Looks also good to me. I cannot test whether the --listen indeed fixes the FTBFS but it shouldn't harm in any case.

PR #167, opened 2021-09-15
One more openssl 3.0 fix

ellert:
Function names are (null) in OpenSSL 3.0 - adapt expected test output.


msalle (approved):
Looks good to me, thanks.

PR #168, opened 2021-09-15
Remove travis secrets for the gctuploader SSH key

matyasselmeci:
This key is not used anymore and is no longer accepted by the repo server.


ellert:
This is a general comment not primarily relevant for this PR only.
Recently PRs end up in the state this PR is currently in stating "Some checks haven’t completed yet", where two test succeed, and the third test (Travis CI - Pull Request) is stuck in the state "Expected — Waiting for status to be reported".
This test never starts, and you have to override the missing test result when pressing the merge button.
Can the test either be removed, or the triggering of the running of the test be fixed so that it runs.

ellert (approved):

brianhlin:
@ellert I've updated the config of the GitHub repo so that PRs require that the two `Build/Test/Deploy` CI tests are required and removed the Travis-CI one. IIRC, we're supposed to still be using Travis for some architectures so it's odd it's not running or otherwise appears to have a communication issue.

fscheiner:
@ellert 
> Recently PRs end up in the state this PR is currently in stating "Some checks haven’t completed yet", where two test succeed, and the third test (Travis CI - Pull Request) is stuck in the state "Expected — Waiting for status to be reported". This test never starts, and you have to override the missing test result when pressing the merge button. Can the test either be removed, or the triggering of the running of the test be fixed so that it runs.

I believe this behaviour is due to the deactivation of Travis-CI builds for PRs I did in August to save build credits. Now that the builds on Travis CI run on ppc64le w/o charge, we can of course reactivate that (=> done now :heavy_check_mark:).

@brianhlin 
> @ellert I've updated the config of the GitHub repo so that PRs require that the two `Build/Test/Deploy` CI tests are required and removed the Travis-CI one. IIRC, we're supposed to still be using Travis for some architectures so it's odd it's not running or otherwise appears to have a communication issue.

So with the current GitHub repo config, does the Travis-CI one still run on merges (and now again on PRs) and is just not required to complete successfully or does it now no longer run at all? If the latter, can you please reconfigure the original settings, because I'd like to keep the non x86 builds?

brianhlin:
@fscheiner I had just changed what CI tests were required to pass before a given PR was allowed to be merged. However, it looks like only the `Travis CI - Pull Request` test is required to pass now and I can't select any `Build/Test/Deploy` GHA as a required test.

fscheiner:
@brianhlin 
> @fscheiner I had just changed what CI tests were required to pass before a given PR was allowed to be merged.

This is done here:
https://github.com/gridcf/gct/settings/branch_protection_rules/1223343

...,right?

> However, it looks like only the `Travis CI - Pull Request` test is required to pass now and I can't select any `Build/Test/Deploy` GHA as a required test.

From https://docs.github.com/en/rest/reference/repos#statuses (linked from the above URL in "_**Require status checks to pass before merging** Choose which [status checks](https://docs.github.com/en/rest/reference/repos#statuses) must pass before branches can be merged..._" ) it sounds like this is for external services ("The status API allows external services ...") only, so maybe this doesn't apply for GitHub actions.

I also found https://github.community/t/github-actions-as-status-checks/16666 which seems to confirm that there is indeed some confusion about GitHub actions not showing up in the branch protection configuration AFAIUI.

Well, we'll see what happens with the next PR.

PR #178, opened 2022-03-06
Sha1 is deprecated

ellert:
This PR addresses test failures on CS9+EPEL9.
CS9 has openssl 3.0.1, which considers sha1 broken and refuses to sign requests that use this algorithm.
The changes are mainly in the generation of test certificates used for running the tests, but there are a few places where the code had to be updated. (This is the reason for the split of the PR in two commits - which should simplify reviewing.)
With these0 changes "make check" succeeds on SC9+EPEL9.


fscheiner (approved):
9d93c6705d0c5130713d07557dd13fb19f80440f looks good to me.

PR #183, opened 2022-04-10
Myproxy ipv6

ellert:
This is a fix for the trust root retrieval feature in the myproxy client.
The PR has two commits. The first has no actual code changes (white space changes only). The second smaler one has the actual changes.


myproxy/source/myproxy.c
```diff
@@ -554,21 +553,16 @@ myproxy_bootstrap_trust(myproxy_socket_attrs_t *attrs)
...
-    char chport[6];
-    snprintf(chport, sizeof(chport), "%d", attrs->psport);
-    BIO_set_conn_port(sbio, chport);
-    BIO_set_conn_hostname(sbio, attrs->pshost);
-    BIO_set_fd(sbio, sockfd, 0);
+    SSL_set_fd(ssl, sockfd);
 
     if (BIO_do_handshake(sbio) <= 0) goto error;
     BIO_write(sbio, "0", 1);    /* GSI deleg flag */
-    if (BIO_flush(sbio) <= 0) goto error;
```
fscheiner, line 571: As the return value of `BIO_write()` is not checked, maybe line 571 (or 566 if kept) was meant to make sure that the flag reaches the other side of the connection.

Or maybe change 570 (now 565) to:
```
if (BIO_write(sbio, "0", 1) < 1) goto error ;    /* GSI deleg flag */
```
...?

ellert: I added a check of the return value of BIO_write as suggested.

ellert:
@fscheiner : I added two more commits to this PR because they are for myproxy too, and if I made a second PR it would conflict with this one since the would both add to the myproxy changlogs in the packaging directory.

myproxy/source/myproxy.c
fscheiner: Thanks.

myproxy/source/ssl_utils.c
```diff
@@ -1274,10 +1282,15 @@ ssl_proxy_store_to_file(SSL_CREDENTIALS         *proxy_creds,
...
         }
     }
 
+    if (tmpfilename)
+    {
+        free(tmpfilename);
+    }
+
     return return_status;
 }
```
fscheiner, line 1283: Good catch!

fscheiner (approved):
Looks fine to me, let's merge this.

PR #192, opened 2022-05-04
Support both SHA1 and SHA256 in myproxy

ellert:
Changing the hash from SHA1 to SHA 256 introduced a backward incompatibility.
OpenSSL 3.0.2 does not allow SHA1, so reverting to SHA1 would not work.
This PR introduces both SHA1 and SHA256 as supported.


msalle (approved):
It looks good to me. We probably should confirm that the failover indeed works as expected (unless you have done that already?)

PR #193, opened 2022-05-07
Can't use non-existing or non-accessible files as source for random data

ellert:
$ grid-cert-request --dir tmp
Enter your name, e.g., John Smith: NN
A certificate request and private key is being created.
You will be asked to enter a PEM pass phrase.
This pass phrase is akin to your account password,
and is used to protect your key file.
If you forget your pass phrase, you will need to
obtain a new certificate.

Can't load /var/adm/wtmp into RNG
80EB3915197F0000:error:12000079:random number generator:RAND_load_file:Cannot open file:crypto/rand/randfile.c:106:Filename=/var/adm/wtmp
Can't load /var/log/messages into RNG
80EB3915197F0000:error:12000079:random number generator:RAND_load_file:Cannot open file:crypto/rand/randfile.c:106:Filename=/var/log/messages
Error number 1 was returned by
   /usr/bin/openssl

### DIFF
--- a/gsi/cert_utils/source/configure.ac
+++ b/gsi/cert_utils/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gsi_cert_utils], [10.9], [https://github.com/gridcf/gct/issues])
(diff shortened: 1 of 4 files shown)


fscheiner (approved):
Looks fine to me!

PR #198, opened 2022-08-05
Remove obsolete GSI-OpenSSH 7.5p1 for Debian

### LINKED ISSUE #170: gsi-openssh for Debian
For 7.5 I managed to compile a debian package for gsi-openssh by omitting part of the patches given in packaging/debian/gsi-openssh/debian/patches/series (because they were redundant anyway). I see those patches are still there and 
make gsi-openssh-deb
still doesn't work straight away.
Maybe someone feels like fixing this..

fscheiner:
Fixes #170


fscheiner:
I worked on a GSI-OpenSSH w/HPN package for Debian GNU/Linux 11 based on the corresponding OpenSSH 8.4p1 since a while and it is available from the OBS now:

https://build.opensuse.org/package/show/home:frank_scheiner:gsi-openssh-debian-11/gsi-openssh ([Repo for Debian](https://download.opensuse.org/repositories/home:/frank_scheiner:/gsi-openssh-debian-11/Debian_11/))

Follow the instructions on https://en.opensuse.org/openSUSE:Build_Service_Debian_builds#Configuring_sources.list to allow installations from this repository. The needed key is https://download.opensuse.org/repositories/home:/frank_scheiner:/gsi-openssh-debian-11/Debian_11/Release.key.

****

So we can now remove the obsolete GSI-OpenSSH 7.5p1 for Debian from the `/packaging/debian` directory. The `/packaging/fedora` directory doesn't contain a RPM spec file for GSI-OpenSSH either. This because we usually want to provide a GSI-OpenSSH version that is based on the OpenSSH version available for the respective OS version.

msalle (commented):
You seem to remove the whole `packaging/debian/gsi-openssh/` directory, is that intended?
Concerning the version, in principle, I presume we no longer need to support 7.5p1, and old-stable has 7.4, not 7.5. Only buster's openssh v1 is 7.5 but I don't think that would be needed.

fscheiner:
> You seem to remove the whole `packaging/debian/gsi-openssh/` directory, is that intended?

Yes, exactly so. That code is unmaintained (i.e. the OpenSSH 7.5p1 source misses any updates or backports for this version, it also does not have any of the Debian patches for this version) and also not really useful anymore. Who would use such an old version voluntarily, if at all it still compiles with newer OpenSSL versions?

> Concerning the version, in principle, I presume we no longer need to support 7.5p1, and old-stable has 7.4

That's already oldoldstable aka Stretch.

>, not 7.5. Only buster's openssh v1 is 7.5 but I don't think that would be needed.

Yeah. And I intentionally started with a GSI-OpenSSH for current Debian stable. I might consider creating GSI-OpenSSH packges for Debian based on OpenSSH 7.9 for oldstable (Buster), though this won't have support for SHA256 and SHA512 based GSSKEX methods AFAIK, same like the GSI-OpenSSH 7.4 in EPEL7.

msalle:
> > You seem to remove the whole `packaging/debian/gsi-openssh/` directory, is that intended?
> 
> Yes, exactly so. That code is unmaintained (i.e. the OpenSSH 7.5p1 source misses any updates or backports for this version, it also does not have any of the Debian patches for this version) and also not really useful anymore. Who would use such an old version voluntarily, if at all it still compiles with newer OpenSSL versions?

ok, my point was more that I wasn't sure it wasn't still used elsewhere but didn't realise we don't yet have gsi-openssh in Debian.

> > Concerning the version, in principle, I presume we no longer need to support 7.5p1, and old-stable has 7.4
> 
> That's already oldoldstable aka Stretch.

again there I had been misreading things, sorry. You're totally right that we shouldn't support stretch

> > , not 7.5. Only buster's openssh v1 is 7.5 but I don't think that would be needed.
> 
> Yeah. And I intentionally started with a GSI-OpenSSH for current Debian stable. I might consider creating GSI-OpenSSH packges for Debian based on OpenSSH 7.9 for oldstable (Buster), though this won't have support for SHA256 and SHA512 based GSSKEX methods AFAIK, same like the GSI-OpenSSH 7.4 in EPEL7.

indeed,that makes perfect sense. Just debian stable for now is fine

msalle (approved):
in short, looks good to me

fscheiner:
> > Yes, exactly so. That code is unmaintained (i.e. the OpenSSH 7.5p1 source misses any updates or backports for this version, it also does not have any of the Debian patches for this version) and also not really useful anymore. Who would use such an old version voluntarily, if at all it still compiles with newer OpenSSL versions?
> 
> ok, my point was more that I wasn't sure it wasn't still used elsewhere but didn't realise we don't yet have gsi-openssh in Debian.

Indeed, there is no GSI-OpenSSH for Debian yet - we had [a discussion about the reasons some time ago](https://mailman.egi.eu/pipermail/discuss/2019-May/000289.html) - and I also checked if it is needed anywhere else in the "standard" GCT build process and it is not AFAICT.  The only thing that won't work after merging ths PR is:

```
make gsi_openssh-deb
```

...but doing that doesn't work out of the box anyhow, because you'd need to build and install the build dependencies first manually.

But with GSI-OpenSSH packages for Debian stable this isn't needed any longer and also not really useful taking into account that each Debian version has its own specific version of OpenSSH.

****

Thanks for having a look. And I'll merge this now.

PR #199, opened 2022-10-28
CI changes

fscheiner:
This PR implements the following changes:

1. Switchover to new source repo hosted by EGI
2. Use Ubuntu Focal (20.04) as build environment for Travis CI instead of the out-of-date but still default Ubuntu Xenial (16.04), clean up the configuration according to config validation at Travis CI and enable builds on ARM64, too. 
3. Add Rocky Linux 9 as build target for GHA and update secret name for uploads to the source repo
4. Build source tarballs and SRPMs for all OS cases for GHA

no.4 unveiled the following issues:

* "globus-xio-udt-driver" can't be built (also for SRPM) on CentOS Stream 9 and Rocky Linux 9 because "libnice(-devel)" is not (yet) available for these OSes, so (S)RPMs are not built for *9 OS versions, too. Mattias has already opened [a bug about branching and building libnice in EPEL 9](https://bugzilla.redhat.com/show_bug.cgi?id=2033520).

* "myproxy-oauth" can't be built on CentOS Stream 8 and Rocky Linux 8 and upwards as it depends on Python 2.x based dependencies not available for the respective OSes, so neither source tarballs nor (S)RPMs are built for *8 and *9 OS versions. There is also no "myproxy-oauth" package in EPEL or Fedora AFAICS. So this part of the GCT could be (1) dropped when switching from CentOS 7 to an *8 version of the OS as main build target or (2) needs to be ported to Python 3 if kept.

* "globus-scheduler-event-generator" depends on LSB (Linux Standard Base) if available (see https://github.com/gridcf/gct/blob/switchover-to-new-src-repo/gram/jobmanager/scheduler_event_generator/source/configure.ac#L59..#L77 for details) but per default does not actually depend on LSB, so for *9 OS versions which are missing "redhat-lsb-core" the LSB dependency can be safely ignored. The dependency comes from [an init script version that uses LSB init functions](https://github.com/fscheiner/gct/blob/switchover-to-new-src-repo/gram/jobmanager/scheduler_event_generator/source/init/globus-scheduler-event-generator-lsb.in#L56). The "globus-scheduler-event-generator-progs" package in EPEL and Fedora use [a non-LSB init script](https://src.fedoraproject.org/rpms/globus-scheduler-event-generator/blob/rawhide/f/globus-scheduler-event-generator) or [a systemd unit file w/o LSB dependency](https://src.fedoraproject.org/rpms/globus-scheduler-event-generator/blob/rawhide/f/globus-scheduler-event-generator%40.service).

### DIFF
--- a/travis-ci/make_rpms.sh
+++ b/travis-ci/make_rpms.sh
@@ -16,6 +16,14 @@ usage()
 
 umask 022
 
+case $(</etc/redhat-release) in
+    CentOS*\ 7*) OS=centos7 ;;
+    CentOS\ Stream*\ 8*) OS=centos-stream-8;;
+    CentOS\ Stream*\ 9*) OS=centos-stream-9;;
+    Rocky\ Linux*\ 8*) OS=rockylinux8 ;;
+    Rocky\ Linux*\ 9*) OS=rockylinux9 ;;
+    *) OS=unknown ;;
+esac
 
 root=$(git rev-parse --show-toplevel)
 packagingdir=$root/packaging
@@ -81,8 +89,27 @@ cat <<EOF >> "$topdir/.rpmmacros"
 %_excludedocs 0
 EOF
 
+# Limit package list according to OS possibilities
+all_packages=( $(grep -v '^#' $fedoradir/ORDERING | grep '[^[:space:]]') )
 
-packages=( $(grep -v '^#' $fedoradir/ORDERING | grep '[^[:space:]]') )
+if [[ $OS == *7 ]]; then
+
+    packages=( ${all_packages[@]} )
+else
+    # myproxy-oauth requires Python 2.x and respective Python 2.x dependencies
+    # that aren't available on CentOS Stream 8 / Rocky Linux 8 and upwards, so
+    # is removed for this case
+    packages=( ${all_packages[@]/myproxy-oauth/} )
+
+    if [[ $OS == *9 ]]; then
+
+        # Not building:
+        # * globus-xio-udt-driver
+        packages_9=( ${packages[@]/globus-xio-udt-driver/} )
+        packages=( ${packages_9[@]} )
+    fi
+fi
+
 
 cp -f "$tarballdir"/*.tar.gz "$topdir"/SOURCES
 
--- a/travis-ci/run_task_inside_docker.sh
+++ b/travis-ci/run_task_inside_docker.sh
@@ -13,6 +13,7 @@ case $(</etc/redhat-release) in
     CentOS\ Stream*\ 8*) OS=centos-stream-8;;
     CentOS\ Stream*\ 9*) OS=centos-stream-9;;
     Rocky\ Linux*\ 8*) OS=rockylinux8 ;;
+    Rocky\ Linux*\ 9*) OS=rockylinux9 ;;
     *) OS=unknown ;;
 esac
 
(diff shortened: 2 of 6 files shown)


travis-ci/run_task_inside_docker.sh
```diff
@@ -78,24 +109,54 @@ elif [[ $TASK == *rpms ]]; then
...
+    else
+        packages+=(redhat-lsb-core)
+    fi
     # for myproxy-oauth
-    packages+=(m2crypto mod_ssl mod_wsgi pyOpenSSL python-crypto)
+    if [[ $OS == *7 ]]; then
+
+        packages+=(m2crypto mod_ssl mod_wsgi pyOpenSSL python-crypto)
+    else
+        # Except for mod_ssl above packages are not available on CentOS Stream
```
msalle, line 148: I guess you would need python3-pyOpenSSL, python3-cryptography, python3-m2crypto and python3-mod_wsgi
?

msalle (commented):
Not sure anyone would still be interested in the myproxy-oauth. In principle https://github.com/ncsa/OA4MP is what's supported, but it still could be nice if it's not too difficult to get to work.

travis-ci/run_task_inside_docker.sh
msalle: I've in the mean time opened PR #200 which might make the code-base of myproxy-oauth use python3

fscheiner: Commented on that in #200.

fscheiner:
> Not sure anyone would still be interested in the myproxy-oauth. In principle https://github.com/ncsa/OA4MP is what's supported, but it still could be nice if it's not too difficult to get to work.

What's the difference between myproxy-oauth and OA4MP anyhow?

msalle:
> > Not sure anyone would still be interested in the myproxy-oauth. In principle https://github.com/ncsa/OA4MP is what's supported, but it still could be nice if it's not too difficult to get to work.
> 
> What's the difference between myproxy-oauth and OA4MP anyhow?

See #200 let's just remove the myproxy-oauth code and close #200.

fscheiner:
Ok, I agree and I'll look into that and come up with another PR.

travis-ci/make_rpms.sh
```diff
@@ -81,8 +89,27 @@ cat <<EOF >> "$topdir/.rpmmacros"
...
+
+if [[ $OS == *7 ]]; then
+
+    packages=( ${all_packages[@]} )
+else
+    # myproxy-oauth requires Python 2.x and respective Python 2.x dependencies
+    # that aren't available on CentOS Stream 8 / Rocky Linux 8 and upwards, so
+    # is removed for this case
+    packages=( ${all_packages[@]/myproxy-oauth/} )
+
```
fscheiner, line 103: That needs to be removed, too.

msalle (approved):
looks good to me

ellert:
> * "globus-xio-udt-driver" can't be built (also for SRPM) on CentOS Stream 9 and Rocky Linux 9 because "libnice(-devel)" is not (yet) available for these OSes, so (S)RPMs are not built for *9 OS versions, too. Mattias has already opened [a bug about branching and building libnice in EPEL 9](https://bugzilla.redhat.com/show_bug.cgi?id=2033520).

Now buildable in EPEL 9. The RHEL 9.1 release added some missing dependencies needed to build the missing dependencies in EPEL.

globus-xio-udt-driver package update for EPEL 9:
https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2022-cd6c9c9c8b

fscheiner:
Thanks for getting this going Mattias (@ellert), I saw it in the morning and will make the required changes.

PR #200, opened 2022-10-30
first try to migrate myproxy-oauth code to python3

msalle:
Mostly automatic changes using 2to3.
Updated RH package names in README.md

### DIFF
--- a/myproxy/oauth/source/README.md
+++ b/myproxy/oauth/source/README.md
@@ -40,8 +40,8 @@ also need libapache2-mod-wsgi and the apache2 server with the ssl module
 enabled and the default-ssl site enabled.
 
 Prerequisite packages on RedHat-based systems
-* pyOpenSSL
-* python-crypto
+* python3-pyOpenSSL
+* python3-pycryptodomex
 
 If you want to use this with apache2 instead of the standalone service, you'll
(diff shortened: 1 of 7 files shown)


fscheiner:
@msalle
Cool, exactly what I had hoped for.

Is there anybody that could confirm that the modified code still works as intended (because there's nothing done for "myproxy-oauth" in the "build/test" step)? Jim (@jbasney), maybe?

When merged, RPM builds on CentOS 7 will stop working, because there the respective Python 3 libs are missing AFAICS. Well, we don't actually build RPMs via the CIs and there's also no "myproxy-oauth" package in EPEL/Fedora, too - so maybe not important.

SRPM builds should work, because they're done with `--nodeps`, but [the installation of prerequsites](https://github.com/gridcf/gct/blob/master/travis-ci/run_task_inside_docker.sh#L92..L93) will fail depending on what packages we intend to install, unless we ignore the CentOS 7 case. As I understand it now, those prerequisite installations are just needed for RPM building, so we could change them as we like, building will just not work everywhere, but that is already the case for *8 and upwards w/o this PR merged.

So should we wait with the merge of this PR until we switch to *8 as main build target?

msalle:
Closing this merge request since we're going to drop the code in any case.

PR #201, opened 2022-10-31
Remove myproxy-oauth

fscheiner:


msalle (approved):
Looks good to me, doesn't seem to be anything of myproxy-oauth left afterwards

ellert (approved):

PR #205, opened 2022-11-28
Fix buffer overflow in test

ellert:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2146585

### DIFF
--- a/gridftp/server/src/test/cmp_alias_ent_test.c
+++ b/gridftp/server/src/test/cmp_alias_ent_test.c
@@ -34,7 +34,7 @@ int main()
         { .alias = "hell", .alias_len = 4 },
         { .alias = (char[]){(char)-1, (char)0 }, .alias_len = 1 },
         { .alias = "hell?", .alias_len = 5 },
-        { .alias = "hell[o]", .alias_len = 5 }
+        { .alias = "hell[o]", .alias_len = 7 }
     };
     test_t tests[] = 
     {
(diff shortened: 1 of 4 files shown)


gridftp/server/src/test/cmp_alias_ent_test.c
```diff
@@ -34,7 +34,7 @@ int main()
         { .alias = "hell", .alias_len = 4 },
         { .alias = (char[]){(char)-1, (char)0 }, .alias_len = 1 },
         { .alias = "hell?", .alias_len = 5 },
-        { .alias = "hell[o]", .alias_len = 5 }
```
fscheiner, line 37: It was in there since originally comitted some 7 years ago. Could be a typo, maybe.

Any idea why this didn't made the test to fail?

fscheiner:
@ellert, @msalle:
So IIUC this `dnf copr enable siddhesh/fortify-source-3` does check the whole source code during built for - let's say obvious - buffer overflow possibilities. Then maybe we should activate that for our CI builds, too, now that we also build RPMs for testing. 

Is something like that also available via EPEL/Fedora repositories?

Some context;
Siddharth Sharma blogged about this on https://www.redhat.com/en/blog/enhance-application-security-fortifysource and this blog post refers to https://gcc.gnu.org/legacy-ml/gcc-patches/2004-09/msg02055.html.

fscheiner (approved):
Looks good to me.

Issue #208, opened 2023-01-16
confusion between ASN1_UTCTIME and ASN1_GENERALIZEDTIME


msalle:
In various places the GCT is confusing `ASN1_UTCTIME` and `ASN1_GENERALIZEDTIME`, assuming `ASN1_TIME` is always a  `ASN1_UTCTIME`.

`globus_gsi_cert_utils_make_time` in [globus_gsi_cert_utils.c](https://github.com/gridcf/gct/blob/master/gsi/cert_utils/source/library/globus_gsi_cert_utils.c) is only valid for `ASN1_UTCTIME`, since it assumes 2-digit years, but is also used to parse any `ASN1_TIME`, for example in myproxy's `ssl_get_times()` and numerous other places. 

In particular [globus_gsi_cert_utils.c lines 245-250](https://github.com/gridcf/gct/blob/master/gsi/cert_utils/source/library/globus_gsi_cert_utils.c#L245-L250) assumes that the year is always 2 digits. That is only true when the type is `V_ASN1_UTCTIME`, i.e. 23. However, from 1/1/2050, it will get encoded as a `V_ASN1_GENERALIZEDTIME` (i.e. 24) and will take 4 digits.

See https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.5 and https://stackoverflow.com/q/10975542 but more in particular https://stackoverflow.com/a/59721373.
The last link also points to [ASN1_TIME_to_tm](https://www.openssl.org/docs/man1.1.1/man3/ASN1_TIME_to_tm.html) which will work from 1.1.1 onwards. Since we still need to support 1.0.2 we cannot use it always unfortunately.

We probably can rework the whole code to use `ASN1_TIME_diff()` instead, which does already work for 1.0.2 and can even directly get the current time.

fscheiner:
Fixed with #209.

fscheiner:
Kept open until next release.

msalle:
@ellert do you think you could add the changes from PR #209 into EPEL and Debian? Alternatively could we make a new release? I again bumped into the issue, this time on a client machine trying to run myproxy-get-delegation which hit the same code. Updating to my temporary RPMs from https://copr.fedorainfracloud.org/coprs/msalle/gct-quickfixes/ fixed it as expected.

fscheiner:
[...]

Issue #207, opened 2023-01-12
Build error: undefined reference to `FIPS_mode'


uihsnv:
When I build the project using `make`, I encounter the following error.

```
/bin/sh ./libtool --mode=link gcc -o ssh ssh.o readconf.o clientloop.o sshtty.o sshconnect.o sshconnect2.o mux.o ssh-sk-client.o -L. -Lopenbsd-compat/  -Wl,-z,relro -Wl,-z,now -Wl,-z,noexecstack -fstack-protector-strong -pie  -lssh -lopenbsd-compat  /home/user/Applications/gct/gsi/gss_assist/source/libglobus_gss_assist.la /home/user/Applications/gct/gsi/gssapi/source/library/libglobus_gssapi_gsi.la /home/user/Applications/gct/common/source/library/libglobus_common.la  -lcrypto -lmd -lz  -lcrypt -lpthread 
[...]

PR #209, opened 2023-01-17
Replace make_time function to work after 2050

msalle:
By using `ASN1_TIME_diff()` instead of manually parsing the data, we make `globus_gsi_cert_utils_make_time()` a lot simpler and also work for `ASN1_GENERALIZEDTIME` and not just `ASN1_UTCTIME` (i.e. it can use `ASN1_TIME`). `ASN1_TIME_diff()` requires OpenSSL >= 1.0.2.
Also rework `globus_gsi_cred_get_lifetime()` to just use `time(NULL)` to get the current UNIX timestamp which means it no longer needs `globus_gsi_cert_utils_make_time()`.
This fixes issue #208

### DIFF
--- a/packaging/fedora/globus-gsi-cert-utils.spec
+++ b/packaging/fedora/globus-gsi-cert-utils.spec
@@ -3,7 +3,7 @@
 Name:		globus-gsi-cert-utils
 %global soname 0
 %global _name %(echo %{name} | tr - _)
-Version:	10.10
+Version:	10.11
 Release:	1%{?dist}
 Summary:	Grid Community Toolkit - Globus GSI Cert Utils Library
 
@@ -180,6 +180,9 @@ make %{?_smp_mflags} check VERBOSE=1
 %doc %{_pkgdocdir}/GLOBUS_LICENSE
 
 %changelog
+* Thu Jan 19 2023 Mischa Salle <msalle@nikhef.nl> - 10.11-1
+- Fix parsing of ASN1 timestamps
+
 * Sat May 07 2022 Mattias Ellert <mattias.ellert@physics.uu.se> - 10.10-1
 - Can't use non-existing or non-accessible files as source for random data
 
--- a/gsi/cert_utils/source/configure.ac
+++ b/gsi/cert_utils/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gsi_cert_utils], [10.10], [https://github.com/gridcf/gct/issues])
+AC_INIT([globus_gsi_cert_utils], [10.11], [https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
 AC_SUBST([MAJOR_VERSION], [${PACKAGE_VERSION%%.*}])
 AC_SUBST([MINOR_VERSION], [${PACKAGE_VERSION##*.}])
--- a/gsi/cert_utils/source/library/globus_gsi_cert_utils.c
+++ b/gsi/cert_utils/source/library/globus_gsi_cert_utils.c
@@ -171,109 +171,42 @@ globus_l_gsi_cert_utils_deactivate(void)
 #endif
 
 /**
- * @brief Convert ASN1_UTCTIME to time_t
+ * @brief Convert ASN1_TIME to time_t
  * @ingroup globus_gsi_cert_utils
  * @details
- * Convert a ASN1_UTCTIME structure to a time_t
+ * Convert a ASN1_TIME structure to a time_t
  *
  * @param ctm
- *        The ASN1_UTCTIME to convert
+ *        The ASN1_TIME to convert
  * @param newtime
  *        The converted time
  *
  * @return
- *        GLOBUS_SUCCESS or an error captured in a globus_result_t
+ *        GLOBUS_SUCCESS or GLOBUS_FAILURE on error
  */
 globus_result_t
 globus_gsi_cert_utils_make_time(
-    const ASN1_UTCTIME *                ctm,
+    const ASN1_TIME *                   ctm,
     time_t *                            newtime)
 {
-    char *                              str;
-    time_t                              offset;
-    char                                buff1[24];
-    char *                              p;
-    int                                 i;
-    struct tm                           tm;
+    int                                 pday, psec;
     globus_result_t                     result;
     static char *                       _function_name_ =
         "globus_gsi_cert_utils_make_time";
 
     GLOBUS_I_GSI_CERT_UTILS_DEBUG_ENTER;
 
-    p = buff1;
-    i = ctm->length;
-    str = (char *)ctm->data;
-    if ((i < 11) || (i > 17))
+    if (ASN1_TIME_diff(&pday, &psec, NULL, ctm))
+    {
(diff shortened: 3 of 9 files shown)


fscheiner (changes requested):
Could you also also update the version numbers and packaging (RPM spec files and debian) in this PR?

msalle:
> Could you also also update the version numbers and packaging (RPM spec files and debian) in this PR?

done, not sure if these are all the files I needed to change. Also, I'm not a maintainer for the debian packages, so Mattias might need to take ownership of the changelog entries to prevent non-maintainer update issues (or make us all debian uploaders).

packaging/fedora/globus-gsi-cert-utils.spec
```diff
@@ -3,7 +3,7 @@
 Name:		globus-gsi-cert-utils
 %global soname 0
```
fscheiner, line 4: @msalle:
I believe we need to increase the `soname` value here as the interface of `globus_gsi_cert_utils_make_time()` changed slightly. Do you agree @ellert?

fscheiner (commented):
Looks good to me*. IIUC with the usage of `time_t` we are nowadays safe beyond 2038 for 64bit architectures (and GNU/Linux: `time_t` is defined as `__TIME64_T_TYPE` in `/usr/include/x86_64-linux-gnu/bitstime64.h` on my Ubuntu 20.04), but 32bit architectures and other OSes (e.g. *BSDs) might not.

****

*) Only for the `soname` value in `packaging/fedora/globus-gsi-cert-utils.spec` I'm not sure about. So let's wait for clarifcation by Mattias (@ellert) before merging this.

fscheiner:
> > Could you also also update the version numbers and packaging (RPM spec files and debian) in this PR?
> 
> done,

Thanks!

> not sure if these are all the files I needed to change. Also, I'm not a maintainer for the debian packages, so Mattias might need to take ownership of the changelog entries to prevent non-maintainer update issues (or make us all debian uploaders).

I think Mattias will adapt these changelogs for the actual Debian/Ubuntu packages anyhow, so we are safe to change them here in our name.

fscheiner:
Btw, looks like we have our first difference between [CentOS Stream 9](https://github.com/gridcf/gct/actions/runs/3959171508/jobs/6781650416) and [Rocky Linux 9](https://github.com/gridcf/gct/actions/runs/3959171508/jobs/6781650851). The build build for CentOS Stream 9 doesn't work w/o #210 merged, the build for Rocky Linux 9 does work. I suspect the Docker images use different versions of OpenSSL, I'll check my VMs for any differences here.

packaging/fedora/globus-gsi-cert-utils.spec
msalle: I considered that too, but `ASN1_TIME` is more generic than `ASN1_UTCTIME` so it's still abi compatible. Furthermore AFAICT `ASN1_TIME *`  and `ASN1_UTCTIME *` are actually pointers to the same struct (see `/usr/include/openssl/ossl_typ.h`) ?

fscheiner: But we can't know if this stays like that. IMHO if the interface changes types, that should be noted, even if they're the same internally, otherwise we would have to check that they stay the same for every new version of OpenSSL.

fscheiner: I mean it might be unlikely, but what is the intention behind giving the same thing multiple names?

msalle:
> `time_t`

considering `time_t` this is a much wider problem all over gct, so we cannot fix that here. I think it is relatively ok nowadays for most common UNIX-likes. See https://en.wikipedia.org/wiki/Year_2038_problem#Implemented_solutions

msalle:
> Btw, looks like we have our first difference between [CentOS Stream 9](https://github.com/gridcf/gct/actions/runs/3959171508/jobs/6781650416) and [Rocky Linux 9](https://github.com/gridcf/gct/actions/runs/3959171508/jobs/6781650851). The build build for CentOS Stream 9 doesn't work w/o #210 merged, the build for Rocky Linux 9 does work. I suspect the Docker images use different versions of OpenSSL, I'll check my VMs for any differences here.

Bit off-topic, but see also https://indico.cern.ch/event/1224843/#12-linux-future-in-wlcg-status i.e. WLCG also seems to be moving away from CentOS Stream in favour of ALMA and Rocky.

fscheiner:
> Bit off-topic, but see also https://indico.cern.ch/event/1224843/#12-linux-future-in-wlcg-status i.e. WLCG also seems to be moving away from CentOS Stream in favour of ALMA and Rocky.

Yeah, I read about that recently in some Linux newsletter.

It's still good that we also build for CentOS Stream 9, because there might not always be a user that stumbles upon a mistake by me (see #207 and #210) and then the failing builds on CentOS Stream 9 would have uncovered it. ;-)

packaging/fedora/globus-gsi-cert-utils.spec
msalle: Just to clarify: it's fine with me either way, I just don't know whether we should. So let's see what @ellert decides. And then one of you feel free to push the corresponding commit (I think you can) if necessary, or I'll do it later.

Btw note that this `soname` in the spec file is only used for SUSE-derived (i.e. when `suse_version` is non-zero). Why isn't that taken directly from the `configure.ac`, i.e. from `MAJOR_VERSION`, `MINOR_VERSION` and `AGE_VERSION`? (see https://github.com/pvanhoof/dir-examples#what-is-right for a very nice explanation of libtool and versioning).

ellert:
I think we should avoid soname changes if possible.

`man ASN1_TIME_set` says:

       The ASN1_TIME structure corresponds to the ASN.1 structure Time defined
       in RFC5280 et al. The time setting functions obey the rules outlined in
       RFC5280: if the date can be represented by UTCTime it is used, else
       GeneralizedTime is used.

       It is recommended that functions starting with ASN1_TIME be used
       instead of those starting with ASN1_UTCTIME or ASN1_GENERALIZEDTIME.
       The functions starting with ASN1_UTCTIME and ASN1_GENERALIZEDTIME act
       only on that specific time format. The functions starting with
       ASN1_TIME will operate on either format.

So changing the argument from `const ASN1_UTCTIME *` to a `const ASN1_TIME *` in globus_gsi_cert_utils_make_time() (which is the only function listed in a header file that is changed) should be fine.

The calls to globus_gsi_cert_utils_make_time() in gsi/callback/source/library/globus_gsi_callback.c already calls it using a `const ASN1_TIME *` pointer anyway...

The same is true for the calls to globus_gsi_cert_utils_make_time() in gsi/credential/source/library/globus_gsi_cred_handle.c  and myproxy/source/ssl_utils.c, since X509_get_notAfter() and X509_get_notBefore() return a `ASN1_TIME *`.

So changing the type in the signature in the header actually reflects how the function is called now.

fscheiner:
Fine with me. :-)

packaging/fedora/globus-gsi-cert-utils.spec
fscheiner: Resolved with https://github.com/gridcf/gct/pull/209#issuecomment-1399230962

ellert (approved):

PR #211, opened 2023-01-31
configure.ac: Avoid calling the undeclared exit function

fweimer-rh:
The SASL and LDAP version checks called exit without including <stdlib.h>.  Future compilers will not support implicit function declarations, causing these two checks to fail unconditionally. Instead, return directly from the main function.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

Alternative to globus/globus-toolkit#132.


msalle (approved):
looks good to me.

PR #213, opened 2023-02-18
Remove register qualifier that is incompatible with C++17

ellert:
~~~
/usr/include/globus/globus_libc.h:206:45: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  206 | int globus_libc_setenv(register const char *name,
      |                                             ^~~~
/usr/include/globus/globus_libc.h:207:45: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  207 |                        register const char *value,
      |                                             ^~~~~
/usr/include/globus/globus_libc.h:209:48: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  209 | void globus_libc_unsetenv(register const char *name);
      |                                                ^~~~
~~~


fscheiner (approved):
Looks good to me.

PR #214, opened 2023-07-12
Add gsi openssh 9.0 from fedora 38 current stable

fscheiner:
The changes are based on:

 https://kojipkgs.fedoraproject.org//packages/gsi-openssh/9.0p1/6.fc38/src/gsi-openssh-9.0p1-6.fc38.src.rpm

...and include a fix to make GSI authentication with gssapi-keyex working again plus a missing change as per https://github.com/openssh/openssh-portable/commit/dbb339f015c33d63484261d140c84ad875a9e548.

****

This PR also includes a change for the CI builds via GHA.


msalle (approved):
Given that it fixes the fedora builds again, I'd say looks good to me!
And the only failing test pipeline seems to have just run out of time, so I think that's fine too.
In general I would probably prefer to have a commit like the first one (which updates to using `actions/checkout@v3`) in a separate PR, since it's unrelated to v9.0 from f38, but I think in this case there is a good reason for combining it.

ellert (approved):

fscheiner:
> In general I would probably prefer to have a commit like the first one (which updates to using `actions/checkout@v3`) in a separate PR, since it's unrelated to v9.0 from f38, but I think in this case there is a good reason for combining it.

I agree, but as it is such a small change, I didn't want to make a separate PR for it. And IIRC it was needed to make the GHA jobs work again. Unfortunately I don't have retained a log that shows this issue.

PR #216, opened 2023-07-31
Checksum error checks fixes #215

### LINKED ISSUE #215: Lack of IO error checks generate incorrect file checksums
The routine [globus_l_gass_copy_cksm_file()](https://github.com/gridcf/gct/blob/7fd95a8262453f5ef274439452c9d4d0989c419b/gass/copy/source/globus_gass_copy_glob.c#L2029) in [globus_gass_copy_glob.c](https://github.com/gridcf/gct/blob/7fd95a8262453f5ef274439452c9d4d0989c419b/gass/copy/source/globus_gass_copy_glob.c)
has a file read loop
`while((n = read(fd,buf,count)) > 0)`
If an IO error occurs in the read call it will return -1 and the loop will terminate early generating an incorrect digest for the file but no error report until the subsequent file transfer fails its checksum test.

On a related note if you check for errors properly you discover that a recursive globus-url-copy attempts to calculate checksums on directory-urls using this routine (which always generates a error in the read call)

My POC fix is
```c
static
globus_result_t
globus_l_gass_copy_cksm_file(
    globus_gass_copy_handle_t *         handle,
    char *                              url,
    char *                              cksm,
    globus_off_t                        offset,
    globus_off_t                        length,
    const char *                        algorithm,
    globus_gass_copy_callback_t         callback,
    void *                              callback_arg)
{
    char *      myname = "globus_l_gass_copy_cksm_file";

    globus_url_t                        parsed_url;
    globus_result_t                     result;
    int                                 rc;

    EVP_MD_CTX *                        mdctx;
    const EVP_MD *                      md;
    char *                              cksmptr;
    unsigned char                       md_value[EVP_MAX_MD_SIZE];
    char                                cksm_buff[CKSM_SIZE];
    unsigned int                        md_len;

    char                                buf[GASS_COPY_CKSM_BUFSIZE];

    int                                 i;
    int                                 fd;
    int                                 n;
    ssize_t                             readlen;
    globus_off_t                        count;
    globus_off_t                        read_left;
    struct stat                         statbuf;
    int                                 is_regular;

    rc = globus_url_parse_loose(url, &parsed_url);
    if(rc != 0)
    {
        result = globus_error_put(
            globus_error_construct_string(
                GLOBUS_GASS_COPY_MODULE,
                GLOBUS_NULL,
                "[%s]: error parsing url: "
                "globus_url_parse returned %d",
                myname,
                url));
        goto error_url;
    }

    if(parsed_url.url_path == GLOBUS_NULL)
    {
        result = globus_error_put(
            globus_error_construct_string(
                GLOBUS_GASS_COPY_MODULE,
                GLOBUS_NULL,
                "[%s]: error parsing url: "
                "url has no path",
                myname));
        goto error_fd;
    }    
       
    read_left = length;
    if(length >= 0)
    {
        count = (read_left > GASS_COPY_CKSM_BUFSIZE) ? 
            GASS_COPY_CKSM_BUFSIZE : read_left;
    }
    else
    {
        count = GASS_COPY_CKSM_BUFSIZE;
    }
    
    fd = open(parsed_url.url_path, O_RDONLY);        
    if(fd < 0)
    {
        result = globus_error_put(
            globus_error_construct_string(
                GLOBUS_GASS_COPY_MODULE,
                GLOBUS_NULL,
                "[%s]: error opening checksum file %s",
                myname,
                parsed_url.url_path));
        goto error_fd;
    }
    fstat(fd,&statbuf);
    is_regular = S_ISREG(statbuf.st_mode);    

    if (is_regular && lseek(fd, offset, SEEK_SET) == -1)
    {
        result = globus_error_put(
            globus_error_construct_string(
                GLOBUS_GASS_COPY_MODULE,
                GLOBUS_NULL,
                "[%s]: error (%d) seeking checksum file %s",
                myname,
                errno,
                parsed_url.url_path));
        goto error_seek;
    }

    OpenSSL_add_all_algorithms();
    md = EVP_get_digestbyname(algorithm);
    if (!md)
    {
        /*
         * try again with uppercase algorithm name.
         */
        char *                          p;
        char *                          alg = strdup(algorithm);
        for(p = alg; *p != '\0'; p++)
        {
            *p = toupper(*p);
        }
        md = EVP_get_digestbyname(alg);
        free(alg);
    }
    if (!md)
    {
        result = globus_error_put(globus_error_construct_string(
                GLOBUS_GASS_COPY_MODULE,
                GLOBUS_NULL, 
                "Unable to use checksum algorithm %s", 
                algorithm));
        goto error_seek;
    }

    mdctx = EVP_MD_CTX_create();
    EVP_DigestInit_ex(mdctx, md, NULL);

    if( is_regular )
    {
        while((n = (readlen = read(fd, buf, count))) > 0)
        {
            if(length >= 0)
            {
                read_left -= n;
                count = (read_left > GASS_COPY_CKSM_BUFSIZE) ? GASS_COPY_CKSM_BUFSIZE : read_left;
            }

            EVP_DigestUpdate(mdctx, buf, n);
        }
        if( readlen < 0 && count > 0 ){
            result = globus_error_put(globus_error_construct_string(
                    GLOBUS_GASS_COPY_MODULE,
                    GLOBUS_NULL,
                    "Read error in checksum calculation url=%s read_left =%d %d %s",
                    url,read_left,readlen, strerror(errno)));
            goto error_seek;
        }
    }

    EVP_DigestFinal_ex(mdctx, md_value, &md_len);
    EVP_MD_CTX_destroy(mdctx);
    
    close(fd);
        
    cksmptr = cksm_buff;
    for (i = 0; i < md_len; i++)
    {
       sprintf(cksmptr, "%02x", md_value[i]);
       cksmptr++;
       cksmptr++;
    }
    *cksmptr = '\0';
    
    strncpy(cksm, cksm_buff, sizeof(cksm_buff));
    //globus_libc_fprintf(stderr,"SPB file-checksum %s=%s\n",url,cksm); 
    globus_url_destroy(&parsed_url);
    
    if(callback)
    {
        callback(callback_arg, handle, NULL);
    }
    return GLOBUS_SUCCESS;

error_seek:
    close(fd);
error_fd:
    globus_url_destroy(&parsed_url);

error_url:

    return result;
}
```
But it probably needs a bit of tidying up.

spbooth:
suggested fix for issue #215

### DIFF
--- a/gass/copy/source/globus_gass_copy_glob.c
+++ b/gass/copy/source/globus_gass_copy_glob.c
@@ -2053,9 +2053,10 @@ globus_l_gass_copy_cksm_file(
 
     int                                 i;
     int                                 fd;
-    int                                 n;
+    ssize_t                             n;
     globus_off_t                        count;
     globus_off_t                        read_left;
+    globus_gass_copy_glob_stat_t        statbuf;
 
     rc = globus_url_parse_loose(url, &parsed_url);
     if(rc != 0)
@@ -2080,44 +2081,7 @@ globus_l_gass_copy_cksm_file(
                 "[%s]: error parsing url: "
                 "url has no path",
                 myname));
-        goto error_fd;
-    }
-
-    read_left = length;
-    if(length >= 0)
-    {
-        count = (read_left > GASS_COPY_CKSM_BUFSIZE) ?
-            GASS_COPY_CKSM_BUFSIZE : read_left;
-    }
-    else
-    {
-        count = GASS_COPY_CKSM_BUFSIZE;
-    }
-
-    fd = open(parsed_url.url_path, O_RDONLY);
-    if(fd < 0)
-    {
-        result = globus_error_put(
-            globus_error_construct_string(
-                GLOBUS_GASS_COPY_MODULE,
-                GLOBUS_NULL,
-                "[%s]: error opening checksum file %s",
-                myname,
-                parsed_url.url_path));
-        goto error_fd;
-    }
-
-    if (lseek(fd, offset, SEEK_SET) == -1)
-    {
-        result = globus_error_put(
-            globus_error_construct_string(
-                GLOBUS_GASS_COPY_MODULE,
-                GLOBUS_NULL,
-                "[%s]: error (%d) seeking checksum file %s",
-                myname,
-                errno,
-                parsed_url.url_path));
-        goto error_seek;
+        goto error_md;
     }
 
     OpenSSL_add_all_algorithms();
@@ -2143,27 +2107,91 @@ globus_l_gass_copy_cksm_file(
             GLOBUS_NULL,
             "Unable to use checksum algorithm %s",
             algorithm));
-        goto error_seek;
+        goto error_md;
     }
 
     mdctx = EVP_MD_CTX_create();
     EVP_DigestInit_ex(mdctx, md, NULL);
 
-    while((n = read(fd, buf, count)) > 0)
+    /*
+     * A recursive copy may attempt to perform a checksum on directory urls.
+     * To preserve behaviour with previous version return a no-data checksum
+     */
+    result = globus_l_gass_copy_stat_file(url, &statbuf);
+    if (result != GLOBUS_SUCCESS)
     {
+        goto error_fd;
+    }
+
+    if (statbuf.type == GLOBUS_GASS_COPY_GLOB_ENTRY_FILE)
+    {
+        read_left = length;
         if(length >= 0)
         {
-            read_left -= n;
-            count = (read_left > GASS_COPY_CKSM_BUFSIZE) ? GASS_COPY_CKSM_BUFSIZE : read_left;
+            count = (read_left > GASS_COPY_CKSM_BUFSIZE) ?
+                GASS_COPY_CKSM_BUFSIZE : read_left;
+        }
+        else
+        {
+            count = GASS_COPY_CKSM_BUFSIZE;
         }
 
-        EVP_DigestUpdate(mdctx, buf, n);
+        fd = open(parsed_url.url_path, O_RDONLY);
+        if (fd < 0)
+        {
+            result = globus_error_put(
+                globus_error_construct_string(
+                    GLOBUS_GASS_COPY_MODULE,
+                    GLOBUS_NULL,
+                    "[%s]: error opening checksum file %s",
+                    myname,
+                    parsed_url.url_path));
+            goto error_fd;
+        }
+
+        if (lseek(fd, offset, SEEK_SET) == -1)
+        {
+            result = globus_error_put(
+                globus_error_construct_string(
+                    GLOBUS_GASS_COPY_MODULE,
+                    GLOBUS_NULL,
+                    "[%s]: error (%d) seeking checksum file %s",
+                    myname,
+                    errno,
+                    parsed_url.url_path));
+            goto error_seek;
+        }
+
+        while ((n = read(fd, buf, count)) > 0)
+        {
+            if (length >= 0)
+            {
+                read_left -= n;
+                count = (read_left > GASS_COPY_CKSM_BUFSIZE) ? GASS_COPY_CKSM_BUFSIZE : read_left;
+            }
+
+            EVP_DigestUpdate(mdctx, buf, n);
+        }
+
+        if (n < 0 && count > 0)
+        {
+            /*
+             * Check for read errors. Otherwise an incorrect checksum is returned
+             * and file-system read errors are mis-reported as corruption
+             */
+            result = globus_error_put(globus_error_construct_string(
+                GLOBUS_GASS_COPY_MODULE,
+                GLOBUS_NULL,
+                "Read error in checksum calculation for file %s read_left=%d count=%d n=%d %s",
+                parsed_url.url_path, read_left, count, n, strerror(errno)));
+            goto error_seek;
+        }
+
+        close(fd);
     }
 
     EVP_DigestFinal_ex(mdctx, md_value, &md_len);
     EVP_MD_CTX_destroy(mdctx);
-
-    close(fd);
 
     cksmptr = cksm_buff;
     for (i = 0; i < md_len; i++)
@@ -2187,8 +2215,9 @@ globus_l_gass_copy_cksm_file(
 error_seek:
     close(fd);
 error_fd:
+    EVP_MD_CTX_destroy(mdctx);
+ error_md:
     globus_url_destroy(&parsed_url);
-
 error_url:
 
     return result;
--- a/gass/copy/source/globus_gass_copy_glob.c
+++ b/gass/copy/source/globus_gass_copy_glob.c
@@ -35,7 +35,7 @@ int globus_l_win_fnmatch(
  ***********************************************************/
 
 
-/*    
+/*
 todo:
 come up with minimal globbing support for windows
 */
@@ -67,7 +67,7 @@ typedef struct
     void *                             entry_user_arg;
     globus_gass_copy_callback_t        op_cb;
     void *                             op_cb_arg;
-} globus_l_gass_copy_glob_info_t; 
+} globus_l_gass_copy_glob_info_t;
 
 static
 globus_result_t
@@ -83,12 +83,12 @@ static
 globus_result_t
 globus_l_gass_copy_glob_parse_ftp_list(
     globus_l_gass_copy_glob_info_t *   info);
-    
+
 static
 globus_result_t
 globus_l_gass_copy_glob_ftp_list(
     globus_l_gass_copy_glob_info_t *   info);
-    
+
 static
 void
 globus_l_gass_copy_ftp_client_op_done_callback(
@@ -96,7 +96,7 @@ globus_l_gass_copy_ftp_client_op_done_callback(
         globus_ftp_client_handle_t *   handle,
         globus_object_t *              err);
 
-static    
+static
 void
 globus_l_gass_copy_ftp_client_list_read_callback(
     void *                             user_arg,
@@ -111,7 +111,7 @@ static
 globus_result_t
 globus_l_gass_copy_mkdir_file(
     char *                              url);
-    
+
 static
 globus_result_t
 globus_l_gass_copy_mkdir_ftp(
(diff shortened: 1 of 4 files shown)


gass/copy/source/globus_gass_copy_glob.c
```diff
@@ -2148,16 +2157,34 @@ globus_l_gass_copy_cksm_file(
...
-
-    while((n = read(fd, buf, count)) > 0)
+    /** A recursive copy may attempt to perform a checksum on directory urls.
+     * To preserve behaviour with previous version return a no-data checksum
+     *
+     */
+    if( is_regular )
     {
-        if(length >= 0)
+        while((n = (readlen = read(fd, buf, count))) > 0)
```
msalle, line 2166: this is a bit of an odd construction: `n` and `readlen` are both set to the return value of read, why do you need both?

gass/copy/source/globus_gass_copy_glob.c
```diff
@@ -2093,33 +2096,39 @@ globus_l_gass_copy_cksm_file(
...
-                GLOBUS_GASS_COPY_MODULE,
-                GLOBUS_NULL,
-                "[%s]: error (%d) seeking checksum file %s",
-                myname,
-                errno,
-                parsed_url.url_path));
-        goto error_seek;
-    }
-
     OpenSSL_add_all_algorithms();
```
msalle, line 2132: Maybe I missed something, but can't we move this whole block, up to line 2150 / 2159? In that case I think you can merge the two `is_regular` blocks into one and don't even need it any more, you can just use the `statbuf.type == GLOBUS_GASS_COPY_GLOB_ENTRY_FILE` as the expression in the single `if`.

gass/copy/source/globus_gass_copy_glob.c
```diff
@@ -2175,7 +2202,7 @@ globus_l_gass_copy_cksm_file(
     *cksmptr = '\0';
     
     strncpy(cksm, cksm_buff, sizeof(cksm_buff));
-    
+    //globus_libc_fprintf(stderr,"SPB file-checksum %s=%s\n",url,cksm);
```
msalle, line 2205: I think we probably shouldn't put (commented-out) debug statements in the code. Although it of course can be very useful.

msalle (changes requested):
I haven't looked at the changes in detail, so I might have missed a few bits. Most important part is of course whether it fixes the issue. In any case I would clean up the code a bit which will make it easier to read. See my inline comments.

gass/copy/source/globus_gass_copy_glob.c
spbooth: probably don't need both.  However n is an int where readlen is the correct typedef for the return value. I wrote this while trying to debug some errors that turned out to be in the file-system and was getting paranoid

ellert (commented):
The changes introduced are not consistent in its use of tabs and spaces.
Before the changes the file used (mostly) spaces only.

The PR has accumulated quite a lot of commits, where some changes things already changed in earlier commits. The PR could benefit from rearranging/merging the commits into logical units (or even merging ito one single commit).

ellert:
@spbooth : I have updated the PR after reviewing it.

There are two changes that are not only white space corrections.
a) I moved the close(fd) statement up into the if(this is a file) conditional, since it is opened inside it.
b) Since the mdctx now lives much longer and there are goto errors while it exists I added its destruction to the error branch and adjusted some of the goto labels.

Let me know if I have misunderstood the intentions of the PR.

fscheiner:
@msalle: Are you OK with the changes made?

gass/copy/source/globus_gass_copy_glob.c
```diff
@@ -2143,21 +2122,62 @@ globus_l_gass_copy_cksm_file(
...
+    		goto error_seek;
+    	}
+        while((n = read(fd, buf, count)) > 0)
         {
-            read_left -= n;
-            count = (read_left > GASS_COPY_CKSM_BUFSIZE) ? GASS_COPY_CKSM_BUFSIZE : read_left;
-        }
+            if(length >= 0)
+            {
+                read_left -= n;
```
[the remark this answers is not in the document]
msalle: This comment still is open and reading it again I'm pretty sure it's really not ok. I know it comes from the original code, but I think it's broken.
Note that `read_left` is set to `length` in line 2086/2087. After that length is only used in the line directly afterwards and here.

msalle (changes requested):
Hi,
there is still one comment open from me, about the use of `length` versus ` read_left`.
I think that needs to be resolved/checked before we can merge it in. I'm pretty sure my suggested change is the correct one. I think other than that it looks ok to me.

gass/copy/source/globus_gass_copy_glob.c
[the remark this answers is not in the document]
fscheiner: @spbooth: Can you comment on the above or already change the PR accordingly?

ellert:
Can we merge this?

msalle:
> Can we merge this?

@ellert What are your thoughts on my comment https://github.com/gridcf/gct/pull/216/files#r1290396767

ellert:
> > Can we merge this?
> 
> @ellert What are your thoughts on my comment https://github.com/gridcf/gct/pull/216/files#r1290396767

I already gave a long reply to this in the discussion above. As far as I can see the code is correct as it is.

msalle:
> > > Can we merge this?
> > 
> > 
> > @ellert What are your thoughts on my comment https://github.com/gridcf/gct/pull/216/files#r1290396767
> 
> I already gave a long reply to this in the discussion above. As far as I can see the code is correct as it is.

Not sure where you mean?
Just to clarify, my question is specifically about the [lines 2165-2169 in globus_gass_copy_glob.c](https://github.com/spbooth/gct/blame/master/gass/copy/source/globus_gass_copy_glob.c#L2165-L2169) in Stephen's fork
Note that `read_left` is set to `length` in [line 2128](https://github.com/spbooth/gct/blame/master/gass/copy/source/globus_gass_copy_glob.c#L2128).
In the 5 lines in the while loop, `n` keeps being subtracted from `read_left` but the check is whether `length` is bigger equal 0 which doesn't seem to make sense unless I completely miss the reasoning? I would expect a check such as `if (read_left >= n)`
The lines are there in the original code too, see [lines 2152-2156 there](https://github.com/gridcf/gct/blame/master/gass/copy/source/globus_gass_copy_glob.c#L2152-L2156) but since we are reordering and in any case fixing a bug, I think we should check this part is correct.

ellert:
> > > > Can we merge this?
> > > 
> > > 
> > > @ellert What are your thoughts on my comment https://github.com/gridcf/gct/pull/216/files#r1290396767
> > 
> > 
> > I already gave a long reply to this in the discussion above. As far as I can see the code is correct as it is.
> 
> Not sure where you mean? Just to clarify, my question is specifically about the [lines 2165-2169 in globus_gass_copy_glob.c](https://github.com/spbooth/gct/blame/master/gass/copy/source/globus_gass_copy_glob.c#L2165-L2169) in Stephen's fork Note that `read_left` is set to `length` in [line 2128](https://github.com/spbooth/gct/blame/master/gass/copy/source/globus_gass_copy_glob.c#L2128). In the 5 lines in the while loop, `n` keeps being subtracted from `read_left` but the check is whether `length` is bigger equal 0 which doesn't seem to make sense unless I completely miss the reasoning? I would expect a check such as `if (read_left >= n)` The lines are there in the original code too, see [lines 2152-2156 there](https://github.com/gridcf/gct/blame/master/gass/copy/source/globus_gass_copy_glob.c#L2152-L2156) but since we are reordering and in any case fixing a bug, I think we should check this part is correct.

Here I quote my reply again. The check is supposed to check if the original value of `length` was negative since this is the condition that determines if the `bytes_left` counter is used or not.

Either `length` is negative, and the routine reads all data until the end of the file, in this case the `bytes_left` counter is not used, and `count` stays constant. Or `length` is non-negative and the routine keeps track of the number of bytes read, i.e. the `bytes_left` counter is used, and `count` is adjusted to never exceed `bytes_left`.

The check `input >= 0` determines whether `bytes_left` and `count` should be adjusted after each call to `read()` or not.

When the `bytes_left` counter is used, `count` is never greater than `bytes_left`. The `n` returned from `read()` is never greater than `count`, so `bytes_left` will never become negative when decreased by `n`, so `count` is never negative.

When the `bytes_left` counter is not used, `count` is never changed, so `count` is never negative.

msalle:
> > > > > Can we merge this?
> > > > 
> > > > 
> > > > @ellert What are your thoughts on my comment https://github.com/gridcf/gct/pull/216/files#r1290396767
> > > 
> > > 
> > > I already gave a long reply to this in the discussion above. As far as I can see the code is correct as it is.
> > 
> > 
> > Not sure where you mean? Just to clarify, my question is specifically about the [lines 2165-2169 in globus_gass_copy_glob.c](https://github.com/spbooth/gct/blame/master/gass/copy/source/globus_gass_copy_glob.c#L2165-L2169) in Stephen's fork Note that `read_left` is set to `length` in [line 2128](https://github.com/spbooth/gct/blame/master/gass/copy/source/globus_gass_copy_glob.c#L2128). In the 5 lines in the while loop, `n` keeps being subtracted from `read_left` but the check is whether `length` is bigger equal 0 which doesn't seem to make sense unless I completely miss the reasoning? I would expect a check such as `if (read_left >= n)` The lines are there in the original code too, see [lines 2152-2156 there](https://github.com/gridcf/gct/blame/master/gass/copy/source/globus_gass_copy_glob.c#L2152-L2156) but since we are reordering and in any case fixing a bug, I think we should check this part is correct.
> 
> Here I quote my reply again. The check is supposed to check if the original value of `length` was negative since this is the condition that determines if the `bytes_left` counter is used or not.
> 
> Either `length` is negative, and the routine reads all data until the end of the file, in this case the `bytes_left` counter is not used, and `count` stays constant. Or `length` is non-negative and the routine keeps track of the number of bytes read, i.e. the `bytes_left` counter is used, and `count` is adjusted to never exceed `bytes_left`.
> 
> The check `input >= 0` determines whether `bytes_left` and `count` should be adjusted after each call to `read()` or not.
> 
> When the `bytes_left` counter is used, `count` is never greater than `bytes_left`. The `n` returned from `read()` is never greater than `count`, so `bytes_left` will never become negative when decreased by `n`, so `count` is never negative.
> 
> When the `bytes_left` counter is not used, `count` is never changed, so `count` is never negative.

Ah, I understand, thanks! 
I must say it's not the most clear code (I would personally have moved the block initialising `read_left` and `count` down to just before the `while` loop, and moved the `if (length>=0)` outside the while loop and combine with the other  `if (length>=0`), but better leave it as is for now.

msalle (approved):

gass/copy/source/globus_gass_copy_glob.c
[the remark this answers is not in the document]
fscheiner: Resolved, see https://github.com/gridcf/gct/pull/216#issuecomment-1824781632 and following comments.

PR #217, opened 2023-08-30
Make library linking survive -Wl,--as-needed

ellert:
The libglobus_rsl_assist library is an empty backwards compatibility library that is there to fulfil link dependencies from earlier versions of GT/GCT. The functionality that used to be provided by this library now is in libglobus_rsl. When linking the libglobus_rsl_assist library it links to libglobus_rsl (-lglobus_rsl), so that loading the empty compatibility library will load the globus_rsl library. However, many Linux distribution by default uses -Wl,--as-needed when linking during package builds (rpm/dpkg). This means that the link from libglobus_rsl_assist to libglobus_rsl is lost. This commit adds a dummy function that references a symbol in the libglobus_rsl library so that the link information is not removed when linking using -Wl,--as-needed.

In addition this commit declares the symbol "lvl" in libglobus_rsl static. This symbol is not in the globus_* namespace and has a very generic name and should therefore not by globally accessible.

### DIFF
--- a/gram/rsl/source/configure.ac
+++ b/gram/rsl/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_rsl],[11.3],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_rsl],[11.4],[https://github.com/gridcf/gct/issues])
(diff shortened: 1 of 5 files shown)


msalle (approved):
Looks good to me.

PR #220, opened 2023-12-13
Ignore unhandled PEM sequences

chrisburr:
I have a PEM file which contains a section denoted by a custom section. All other software I've tried happily ignores this section however GCT fails with:

```
Unable to read credential for import
globus_gsi_gssapi: Error with GSI credential
globus_credential: Error reading proxy credential: Unhandled PEM sequence: ...
```

[RFC 7468, Section 2](https://www.rfc-editor.org/rfc/rfc7468#section-2) doesn't explicitly state that parsers should ignore custom labels in PEM files, it does imply a level of flexibility in handling unrecognized or non-standard data, suggesting that well-designed parsers may indeed ignore custom labels and proceed with processing the rest of the file.

What do you think of removing this check so that unknown labels are ignored?

### DIFF
--- a/gsi/credential/source/library/globus_gsi_credential.c
+++ b/gsi/credential/source/library/globus_gsi_credential.c
@@ -1025,14 +1025,6 @@ globus_gsi_cred_read_proxy_bio(
                 goto exit;
             }
         }
-        else
-        {
-            GLOBUS_GSI_CRED_OPENSSL_ERROR_RESULT(
-                result,
-                GLOBUS_GSI_CRED_ERROR_READING_PROXY_CRED,
-                (_GCRSL("Unhandled PEM sequence: %s"), name));
-            goto exit;
-        }
         if (save_data)
         {
             OPENSSL_free(save_data);


fscheiner:
Not sure. But if we can safely assume that ignoring it does no harm and that it really is not touched anywhere else, it could be a solution. Maybe a command line switch or environment variable to enable or disable the behaviour would be even better.

fscheiner:
> I have a PEM file which contains a section denoted by a custom section. All other software I've tried happily ignores this section 

What other software did you try?

chrisburr:
> Not sure. But if we can safely assume that ignoring it does no harm and that it really is not touched anywhere else, it could be a solution. Maybe a command line switch or environment variable to enable or disable the behaviour would be even better.

This feels excessive. The RFC is clear that malformed data should be tollerated and the rest of the PEM data results in a valid chain that can be used like normal.

> What other software did you try?

I know these have some overlap in implementation but:

curl, XRootD, davix, gfal2 (except for srm which fails due to gct), DIRAC, openssl

maarten-litmaath:
Hi all,
I think it is OK to skip over any unexpected data and just parse what looks to be the PEM blob.

That is not going to open the door further for anything: the PEM blob still has to be valid.

msalle:
Hi @chrisburr all, 
could you send us the pem file, it would be good for testing and for future reference (I presume it's a certificate not a private key). If you prefer, you can also send it by email to tech(at)gridcf.org.

PR #222, opened 2024-01-18
Type errors

ellert:
Fedora 40 updated gcc to version 14. This version is more strict with assignments between different pointer types. This PR fixes the compilation errors in myproxy due to this change.

The PR has two commits. The first has no code changes, only changes to tabs and spaces. The second commit has the actual code changes.


msalle (approved):
Looks good to me, thank you!

PR #223, opened 2024-03-08
64 bit time_t

ellert:
Debian recently migrated to using a 64 bit time_t on most of their 32 bit architectures.
This leads to issues where a time_t variable is used in printf or sscanf.
This PR addresses those issues (and a few cases with size_t vs. long that I found when checking logs on 32 bit).

### DIFF
--- a/gass/transfer/source/library/globus_gass_transfer_http.c
+++ b/gass/transfer/source/library/globus_gass_transfer_http.c
@@ -314,8 +314,8 @@ globus_l_gass_transfer_http_send(
 
 	/* send chunk header and footer as an iovec array */
 	sprintf((char *) new_proto->iov[0].iov_base,
-		"%lx%s",
-		(long) new_proto->user_buflen,
+		"%zx%s",
+		new_proto->user_buflen,
 		CRLF);
 	new_proto->iov[0].iov_len = strlen((char *) new_proto->iov[0].iov_base);
 
@@ -1728,7 +1728,7 @@ globus_l_gass_transfer_http_request_refer(
 		      GLOBUS_L_HTML_HEADER);
     offset += sprintf(referral_string + offset,
 		      GLOBUS_L_CONTENT_LENGTH_HEADER,
-		      (long) body_count);
+		      body_count);
     offset += sprintf(referral_string + offset,
 		      CRLF);
 
@@ -1838,7 +1838,7 @@ globus_l_gass_transfer_http_request_deny(
 		      GLOBUS_L_HTML_HEADER);
     offset += sprintf(deny_string + offset,
 		      GLOBUS_L_CONTENT_LENGTH_HEADER,
-		      (long) body_count);
+		      body_count);
     offset += sprintf(deny_string + offset,
 		      CRLF);
 
@@ -1941,7 +1941,7 @@ globus_l_gass_transfer_http_request_authorize(
 	{
 	    offset += sprintf(authorize_string + offset,
 			      GLOBUS_L_CONTENT_LENGTH_HEADER,
-			      (long) length);
+			      length);
 	}
 	offset += sprintf(authorize_string + offset,
 			  CRLF);
--- a/gass/transfer/source/configure.ac
+++ b/gass/transfer/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gass_transfer],[9.4],[https://github.com/gridcf/gct/issues])
+AC_INIT([globus_gass_transfer],[9.5],[https://github.com/gridcf/gct/issues])
 AC_CONFIG_MACRO_DIR([m4])
 AC_SUBST([MAJOR_VERSION], [${PACKAGE_VERSION%%.*}])
 AC_SUBST([MINOR_VERSION], [${PACKAGE_VERSION##*.}])
@@ -25,6 +25,6 @@ AC_CONFIG_FILES(
         globus-gass-transfer.pc
         Makefile
         library/Doxyfile
-	library/Makefile
-	version.h)
+        library/Makefile
+        version.h)
 AC_OUTPUT
--- a/gass/transfer/source/library/globus_l_gass_transfer_http.h
+++ b/gass/transfer/source/library/globus_l_gass_transfer_http.h
@@ -128,7 +128,7 @@ static char * globus_l_gass_transfer_http_subject_name;
 
 #define GLOBUS_L_DEFAULT_DENIAL_MESSAGE	"Internal Server Error"
 
-#define GLOBUS_L_CONTENT_LENGTH_HEADER	"Content-Length: %ld" CRLF
+#define GLOBUS_L_CONTENT_LENGTH_HEADER	"Content-Length: %zu" CRLF
 #define GLOBUS_L_CHUNKED_HEADER		"Transfer-Encoding: chunked" CRLF
 #define GLOBUS_L_BINARY_HEADER		"Content-Type: " \
 					    "application/octet-stream" CRLF
--- a/gram/jobmanager/lrms/fork/source/configure.ac
+++ b/gram/jobmanager/lrms/fork/source/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.60])
 
-AC_INIT([globus_gram_job_manager_fork],[3.3],[https://github.com/gridcf/gct/issues])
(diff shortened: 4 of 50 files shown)


ellert:
This took a few iterations to find all corner cases, but I think it is complete now. At least there are no more compiler warnings related to the change, and the tests succeed on all debian architectures for all packages.

gass/transfer/source/library/globus_gass_transfer_http.c
```diff
@@ -4653,7 +4653,7 @@ globus_l_gass_transfer_http_construct_request(
 
 	    sprintf(cmd + strlen(cmd),
 		    GLOBUS_L_CONTENT_LENGTH_HEADER,
-		    length);
+		    (long) length);
```
msalle, line 4656: not sure this is safe: `length` is a `globus_size_t` which is `size_t` which is unsigned plus it might be bigger than an signed long (and is actually a long long on some platforms), meaning it could overflow when cast to long: i.e. `(long)length` could be negative, meaning the number of digits in lines 4633-4637 above isn't taking into account a leading minus sign?

gass/transfer/source/library/globus_gass_transfer_http.c
```diff
@@ -4743,7 +4743,7 @@ globus_l_gass_transfer_http_construct_request(
 		    proto->url.host);
 	    sprintf((char *) cmd + strlen(cmd),
 		    GLOBUS_L_CONTENT_LENGTH_HEADER,
-		    length);
+		    (long) length);
```
msalle, line 4746: see comment above on line 4656

msalle (changes requested):
Other than my two comments I think it looks all fine to me.

gass/transfer/source/library/globus_gass_transfer_http.c
johnny-mnemonic: @ellert, @msalle: Sorry, totally forgot about that on the weekend.

I tried to find out what the exact definition of `size_t` is on my local machine (x86_64 based), it's defined in `stddef.h` (`/usr/lib/gcc/x86_64-linux-gnu/9/include/stddef.h` on my machine) like so:

```
[...]
#ifndef __SIZE_TYPE__
#define __SIZE_TYPE__ long unsigned int
#endif
#if !(defined (__GNUG__) && defined (size_t))
typedef __SIZE_TYPE__ size_t;
[...]
```
Using the following command (from https://unix.stackexchange.com/a/451235) confirms that:
```
$ echo '#include <stddef.h>' | gcc -xc -E - | grep size_t
typedef long unsigned int size_t;
```
And looking into for example http://ftp.de.debian.org/debian/pool/main/g/gcc-12/libgcc-12-dev_12.2.0-14_i386.deb shows that it is also defined like that on i386 (i686 actually).

So it's already a `long` data type. Would the cast to `long` then make any difference?

fscheiner: @ellert, @msalle: Sorry, totally forgot about that on the weekend.

I tried to find out what the exact definition of `size_t` is on my local machine (x86_64 based), it's defined in `stddef.h` (`/usr/lib/gcc/x86_64-linux-gnu/9/include/stddef.h` on my machine) like so:

```
[...]
#ifndef __SIZE_TYPE__
#define __SIZE_TYPE__ long unsigned int
#endif
#if !(defined (__GNUG__) && defined (size_t))
typedef __SIZE_TYPE__ size_t;
[...]
```
Using the following command (from https://unix.stackexchange.com/a/451235) confirms that:
```
$ echo '#include <stddef.h>' | gcc -xc -E - | grep size_t
typedef long unsigned int size_t;
```
And looking into for example http://ftp.de.debian.org/debian/pool/main/g/gcc-12/libgcc-12-dev_12.2.0-14_i386.deb shows that it is also defined like that on i386 (i686 actually).

So it's already a `long` data type. Would the cast to `long` then make any difference?

msalle: Hi Frank,
two points:
1) your test is linux specific, not sure if it's generic on all platforms (I'm pretty sure there are platforms it's a long long).
2) more importantly: casting an *unsigned* long to a *signed* long can swap the sign if it's bigger than half the MAX

fscheiner: But is `long unsigned int` equal to `unsigned long`?

msalle: yes, they are identical. long unsigned and unsigned long can be swapped and long int is the same as long.

fscheiner: Ok, thanks for clarification.

fscheiner: I wonder how Debian handles `size_t` in their t64 transition. Need to check a few package sources maybe...

ellert: The GLOBUS_L_CONTENT_LENGTH_HEADER used here is defined as:

https://github.com/gridcf/gct/blob/e479e5165becde3f6d7cd64e072b8845ca8ad306/gass/transfer/source/library/globus_l_gass_transfer_http.h#L131

I.e. it uses %ld, and therfore expects a signed long. Whether the value is printed as signed or unsigned depends on the format string, not the variable type.
```
$ cat print.c
#include <stdio.h>
int main() {
  long l = -1;
  printf ("%ld\n", l);
  printf ("%lu\n", l);
}
$ gcc -o print print.c 
$ ./print
-1
18446744073709551615
```
So if the size_t is cast to a (signed) long or an unsigned long doesn't matter, it will be printed as a signed long in either case, because that is what the format string says.

The macro GLOBUS_L_CONTENT_LENGTH_HEADER is only used in this file. It is used 5 times. The first 3 already uses a cast to long. This commit adds the same cast to the remaining 2.

An alternative fix would be to do it the opposite way: Redefine the macro to use %zu instead of %ld and remove the 3 existing casts.

On 64 bit Linux size_t is a long.
On 32 bit Linux size_t is an int. Which is why it gives a compiler warning since it doesn't match the %ld format string. int and long are both 32 bits here, but it still complains.

On 64 bit windows a size_t is a long long, because long on 64 bit windows is 32 bit. But I don't know if this is relevant.

ellert: > An alternative fix would be to do it the opposite way: Redefine the macro to use %zu instead of %ld and remove the 3 existing casts.

I have implemented this.

ellert: > ... that it is also defined like that on i386 (i686 actually).
size_t is not the same for x86_64 and ix86:
```
$ echo '#include <stddef.h>' | gcc -xc -E - | grep size_t
typedef long unsigned int size_t;
$ echo '#include <stddef.h>' | gcc -m32 -xc -E - | grep size_t
typedef unsigned int size_t;
```

ellert: > I wonder how Debian handles `size_t` in their t64 transition. Need to check a few package sources maybe...

There are no change to size_t due to the changes in time_t.
There is potentially a change to off_t, since it is not possible to use 64 bit time_t without also using 64 bit off_t. GCT is already using 64 bit off_t (-D_FILE_OFFSET_BITS=64) on 32 bit architectures. So enabling 64 bit time_t doesn't change off_t for GCT.

fscheiner: > > ... that it is also defined like that on i386 (i686 actually).
> > size_t is not the same for x86_64 and ix86:
> 
> ```
> $ echo '#include <stddef.h>' | gcc -xc -E - | grep size_t
> typedef long unsigned int size_t;
> $ echo '#include <stddef.h>' | gcc -m32 -xc -E - | grep size_t
> typedef unsigned int size_t;
> ```

Thanks for checking, I only looked into the definition in the compiler dev package on x86. I assume `__SIZE_TYPE__` or even `size_t` is then already defined differently on x86 by libc or kernel before the default definition (`long unsigned int`) could come into effect.

msalle: > > An alternative fix would be to do it the opposite way: Redefine the macro to use %zu instead of %ld and remove the 3 existing casts.
> 
> I have implemented this.

thanks, that indeed seems like the right fix!

msalle (approved):
I agree with the new version, all looks fine to me

PR #227, opened 2024-07-11
Add openssl-devel-engine build requirement on Fedora 41+

ellert:
The ENGINE API is deprecated in openssl since version 3.0. Fedora is trying to phase it out, and as a first step they have moved the associated header files into a separete package. For now the ENGINE API can still be used if this package is added to the build requriements for the package, as this commit does. Developers are encouraged to use the PROVIDER API instead.

https://fedoraproject.org/wiki/Changes/OpensslDeprecateEngine

### DIFF
--- a/packaging/fedora/myproxy.spec
+++ b/packaging/fedora/myproxy.spec
@@ -3,7 +3,7 @@
 Name:           myproxy
 %global soname 6
 Version:        6.2.17
-Release:        1%{?dist}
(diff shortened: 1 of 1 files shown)


johnny-mnemonic (approved):
Looks good.

fscheiner:
@ellert:
Before merging, let me fix our CIs so the badges get green after the next build.

PR #232, opened 2024-10-30
Compatibility with autoconf 2.72

ellert:
Fix for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1086241


fscheiner:
Looks good to me. Thanks for fixing, @ellert.

I only wonder what changed outside of the GCT that the old code now made a problem.

ellert:
> I only wonder what changed outside of the GCT that the old code now made a problem.

I had a closer look, and this is due to changes in autoconf 2.72.
I have therefore reworked the pull request. Please review again.

fscheiner (approved):
Assuming that (i.e. `[...]file_offset_bits`) is no longer relevant for autoconf 2.72, I'd say let's do this.

fscheiner:
Also the failing check should be fixed by #233.

PR #234, opened 2024-11-02
Make OpenSSL engine optional

ellert:
OpenSSL in RHEL 10 has dropped the support for the ENGINE API.
This PR makes the use of the ENGINE API optional in myproxy. so that it can be compiled on RHEL+EPEL 10.

### DIFF
--- a/myproxy/source/myproxy_server.c
+++ b/myproxy/source/myproxy_server.c
@@ -287,10 +287,17 @@ main(int argc, char *argv[])
     }
 
     if(server_context->certificate_openssl_engine_id) {
+#ifndef OPENSSL_NO_ENGINE
         if(!initialise_openssl_engine(server_context)) {
             myproxy_log_verror();
             my_failure("Could not initialise OpenSSL engine.");
         }
+#else
+        myproxy_log("Openssl has no engine support.");
+        myproxy_log("Can not use certificate_openssl_engine_id option.");
+        myproxy_log("Exiting.");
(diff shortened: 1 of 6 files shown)


myproxy/source/myproxy_server.c
```diff
@@ -287,10 +287,17 @@ main(int argc, char *argv[])
...
 
     if(server_context->certificate_openssl_engine_id) {
+#ifndef OPENSSL_NO_ENGINE
         if(!initialise_openssl_engine(server_context)) {
             myproxy_log_verror();
             my_failure("Could not initialise OpenSSL engine.");
         }
+#else
+        myproxy_log("Openssl has no engine support.");
+        myproxy_log("Can not use certificate_openssl_engine_id oprion.");
```
fscheiner, line 297: Small typo: "oprions" => "options"

fscheiner (approved):
If it doesn't hinder its main functionality, let's do that.

fscheiner:
Context: _"The ENGINE API was introduced in OpenSSL version 0.9.6 as a low level interface for adding alternative implementations of cryptographic primitives, most notably for integrating hardware crypto devices."_

from https://github.com/openssl/openssl/blob/master/README-ENGINES.md

myproxy/source/myproxy_server.c
ellert: Thanks. Fixed.

PR #236, opened 2024-12-26
Merge flags for OpenSSL 3.x versions.

### LINKED ISSUE #235: gsi-openssh configure fails on openssl 3.4.*
OpenSSL 3.4 has been released and seems to work fine with the gct but the configure.ac for gsi-openssh needs to be updated to include commit https://github.com/openssh/openssh-portable/commit/2eded551ba96e66bc3afbbcc883812c2eac02bd7 from upstream openssh (included in 9.8p1 and 9.9p1).
See in particular https://github.com/gridcf/gct/blob/master/gsi_openssh/source/configure.ac#L2906-L2913

msalle:
Taking over upstream openssh commit 2eded551ba96e66bc3afbbcc883812c2eac02bd7:

> OpenSSL has moved to 3.4 which we don't currently accept.  Based on the OpenSSL versioning policy[0] it looks like all of the 3.x versions should work with OpenSSH, so remove the distinction in configure and accept all of them.
> 
> [0] https://openssl.org/policies/general/versioning-policy.html

Fixes #235

### DIFF
--- a/gsi_openssh/source/configure.ac
+++ b/gsi_openssh/source/configure.ac
@@ -2903,12 +2903,8 @@ if test "x$openssl" = "xyes" ; then
 				;;
 			101*)   ;; # 1.1.x
 			200*)   ;; # LibreSSL
-			300*)
+			30*)
 				# OpenSSL 3; we use the 1.1x API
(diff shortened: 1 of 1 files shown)


fscheiner (approved):
Looks fine! Let's merge that...

PR #237, opened 2025-01-17
Fix compilation with GCC 15 (Fedora 42)

ellert:
GCC 15 has improved diagnostics and detects more incompatible pointer type errors than earlier versions. Especially function pointers being assigned pointers to function with the wrong type and/or number of arguments are better detected.

This commit fixes the compilation with GCC 15 on Fedora 42.


fscheiner:
Great! Thanks for fixing that!

fscheiner (approved):
Looks good. Let's do that.

PR #238, opened 2025-04-26
Only check write permission for stdout/stderr if path is absolute

ellert:
For relative paths stdout/stderr are relative to the job's working directory which is writable, so the check is redundant.

The current check instead tests if the user's home directory is writable which is irrelevant. The test also creates and leaves behind a zero-length file in the user's home directory. The test also gives false failures if the user's home directory is not writable.

Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1103552


msalle (approved):
Looks good to me, thanks!

PR #244, opened 2026-04-12
Fix CVE-2026-3497

fscheiner:
Fix information disclosure or denial of service due to uninitialized variables in gssapi-keyex

Based on https://src.fedoraproject.org/rpms/openssh/c/f595270c4ac496ad743336e80a4f3cc11a144efb?branch=rawhide by Zoltan Fridrich <zfridric@redhat.com>

### DIFF
--- a/gsi_openssh/source/kexgsss.c
+++ b/gsi_openssh/source/kexgsss.c
@@ -132,7 +132,7 @@ kexgss_final(struct ssh *ssh,
 {
 	struct kex *kex = ssh->kex;
 	Gssctxt *gss = kex->gss;
-	gss_buffer_desc msg_tok;
+	gss_buffer_desc msg_tok = GSS_C_EMPTY_BUFFER;;
 	u_char hash[SSH_DIGEST_MAX_LENGTH];
 	size_t hashlen;
 	struct sshbuf *shared_secret = NULL;
@@ -217,7 +217,8 @@ input_kexgss_init(int type,
 	Gssctxt *gss = kex->gss;
 	struct sshbuf *empty;
 	struct sshbuf *client_pubkey = NULL;
-	gss_buffer_desc recv_tok, send_tok = GSS_C_EMPTY_BUFFER;
+	gss_buffer_desc recv_tok = GSS_C_EMPTY_BUFFER;
+	gss_buffer_desc send_tok = GSS_C_EMPTY_BUFFER;
 	OM_uint32 ret_flags = 0;
 	int r;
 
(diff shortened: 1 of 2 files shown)


fscheiner:
Untested so far. Also still unclear if the GSI functionality is affected.

gsi_openssh/source/kexgsss.c
```diff
@@ -526,8 +528,7 @@ input_kexgssgex_groupreq(int type,
 
 	kex->dh = mm_choose_dh(min, nbits, max);
 	if (kex->dh == NULL) {
-		sshpkt_disconnect(ssh, "Protocol error: no matching group found");
-		fatal("Protocol error: no matching group found");
+		ssh_packet_disconnect(ssh, "Protocol error: no matching group found");
```
fscheiner, line 531: @msalle : Here I wonder if that is enough, as the original code did end - I guess the key exchange - with fatal() at this point. But [the change by Red Hat](https://src.fedoraproject.org/rpms/openssh/c/f595270c4ac496ad743336e80a4f3cc11a144efb?branch=rawhide) has it that way.

msalle: I'd stick to what RH is doing, although it might look a bit odd.
However from https://github.com/openssh/openssh-portable/blob/master/packet.c#L2121 and further it is clear that `ssh_packet_disconnect()` never returns and so there should indeed be no a reason for a subsequent fatal().

msalle (approved):
I haven't tested it, but I'd say if it builds ok and you have tested also run-time that it works, and given that it's mostly (if not all) the fixes from RH in any case, I'd say it looks fine to me.

gsi_openssh/source/kexgsss.c
fscheiner: Ok. Thanks for looking into it.

fscheiner:
It works the same as before (here tested on Rocky Linux 8.x):
```
[johndoe@gridftp-5 gsi-openssh-testing-with-cve-fix]$ sudo ~/bin/test-gss-kex-for-gsi-openssh.bash gridftp-5.domain.tld johndoe2
gsisshd: OpenSSH_10.0p2c-GSI_hpn18.7.0 GSI, OpenSSL 1.1.1k  FIPS 25 Mar 2021
gsissh: OpenSSH_10.0p2c-GSI_hpn18.7.0 GSI, OpenSSL 1.1.1k  FIPS 25 Mar 2021

Wait 3 seconds for startup of gsisshd ...

gss-gex-sha1- OK ( plus johndoe => johndoe2 )
gss-group1-sha1- OK ( plus johndoe => johndoe2 )
gss-group14-sha256- OK ( plus johndoe => johndoe2 )
gss-nistp256-sha256- OK ( plus johndoe => johndoe2 )
gss-curve25519-sha256- OK ( plus johndoe => johndoe2 )
gss-group16-sha512- OK ( plus johndoe => johndoe2 )
```

...so will merge this now.

PR #245, opened 2026-05-19
Openssl 4.0

### LINKED ISSUE #246: Build failure with OpenSSL 4.0
This bug was filed using automation as part of the Debian OpenSSL 4.0 migration effort. Please ignore if the build issue is already fixed.

The following Debian packages from this repository fail to build against OpenSSL 4.0:

- globus-gridmap-eppn-callout: https://bugs.debian.org/1138406
- globus-gsi-callback: https://bugs.debian.org/1138455
- globus-gsi-cert-utils: https://bugs.debian.org/1138437
- globus-gsi-credential: https://bugs.debian.org/1138320
- globus-gsi-proxy-core: https://bugs.debian.org/1138435
- globus-gsi-proxy-ssl: https://bugs.debian.org/1138372
- globus-gssapi-gsi: https://bugs.debian.org/1138350
- globus-proxy-utils: https://bugs.debian.org/1138392
- myproxy: https://bugs.debian.org/1138316

ellert:
Fixes for compiling with OpenSSL 4.
The first commit only changes tabs and spaces (`git diff --ignore-all-space` is empty). It just fixes inconsistent whitespace in the files that are modified in later commits.
The secons commit contains the changes for OpenSSL 4.
The third commit fixes some compiler and doxygen warnings unrelated to the UpenSSL 4 update.
The last commit contains the version increase andcorresponding packaging file update.


maarten-litmaath:
Hi Mattias,
thanks for that already!

Was anything also tested after those changes?

ellert:
> Was anything also tested after those changes?

I didn't do any additional testing other than running all the tests in `make check`.

Most of the changes are related to replacing direct access to the fields in ASN1_STRING (.data, .length, .type) with calls to ASN1_STRING_get0_data(), ASN1_STRING_length() and ASN1_STRING_type() since the type has been made opaque. And similarly using ASN1_STRING_set() to change the string instead of changing the fields directly.

Other changes are declaring some things const to match changes in the OpenSSL headers, in a few instances conditionaly depending on the OpenSSL version where needed.

msalle:
Hi Mattias,
many thanks! I will try to test it in the coming days and let you know if I see any problems.

msalle:
> Hi Mattias, many thanks! I will try to test it in the coming days and let you know if I see any problems.

I had some time in between and did some build and run-time checks using LCMAPS, which tests a few different globus libraries, see below.
Only real problem I found is that you need to add an entry for 40*) in https://github.com/ellert/gct/blob/openssl-4.0/gsi_openssh/source/configure.ac#L3076. Other than that a lot of compiler warnings, but nothing breaking.

Globus packages tested by the lcas-lcmaps-gt4-interface:
```
libglobus_callout.so.0
libglobus_common.so.0
libglobus_gridmap_callout_error.so.0
libglobus_gsi_callback.so.0
libglobus_gsi_cert_utils.so.0
libglobus_gsi_credential.so.1
libglobus_gsi_proxy_core.so.0
libglobus_gsi_sysconfig.so.1
libglobus_gssapi_error.so.2
libglobus_gssapi_gsi.so.4
libglobus_gss_assist.so.3
libglobus_oldgaa.so.0
libglobus_openssl_error.so.0
libglobus_openssl.so.0
libglobus_proxy_ssl.so.1
```
and in a different setup, using a run used by lcmaps in e.g. gLExec and the like (via llrun):
```
libglobus_common.so.0
libglobus_gsi_callback.so.0
libglobus_gsi_cert_utils.so.0
libglobus_gsi_credential.so.1
libglobus_gsi_proxy_core.so.0
libglobus_gsi_sysconfig.so.1
libglobus_gssapi_gsi.so.4
libglobus_oldgaa.so.0
libglobus_openssl_error.so.0
libglobus_openssl.so.0
libglobus_proxy_ssl.so.1
liblcmaps_return_account_from_pem.so
```

ellert:
Bugs have been filed in Debian:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138316 myproxy
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138320 globus-gsi-credential
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138350 globus-gssapi-gsi
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138372 globus-gsi-proxy-ssl
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138392 globus-proxy-utils
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138406 globus-gridmap-eppn-callout
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138435 globus-gsi-proxy-core
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138437 globus-gsi-cert-utils
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138455 globus-gsi-callback

fscheiner:
@msalle , @ellert :
I added a patch to also accept OpenSSL 4.x when configuring GSI-OpenSSH.

**But GSI-OpenSSH won't build with OpenSSL 4.x, because of the use of `EVP_CIPHER_meth_free()` (and other `EVP_CIPHER_meth_[...]()` functions) in the HPN-SSH patchset that we include.**

@rapier1: Do you plan to support HPN-SSH builds with OpenSSL 4.x in the near future? :-)

tl;dr

Then I wanted to build our in-tree GSI-OpenSSH 10.0 on Debian Sid against OpenSSL 4.x to test its workings. But it looks like this needs some additional preprocessor configurations to allow building on Debian (e.g. there's no `openssl/fips.h` there). Looking into `openssl/fips.h` in Rocky Linux 8, there's `OPENSSL_FIPS` we can check for, to include that FIPS header or not, like so:
```c
#ifdef OPENSSL_FIPS
#include <openssl/fips.h>
#endif
```
...for example. This solves one issue with Debian. But there's also at least one issue with OpenSSL 4.x and the HPN patches - `EVP_CIPHER_meth_free()` (and other EVP_CIPHER_meth_[...]() functions) is/are no longer there in OpenSSL 4.x it seems:
```
cipher.c: In function ‘cipher_free’:
cipher.c:575:17: error: implicit declaration of function ‘EVP_CIPHER_meth_free’; did you mean ‘EVP_CIPHER_CTX_free’? [-Wimplicit-function-declaration]
  575 |                 EVP_CIPHER_meth_free((void *)(EVP_CIPHER *)cc->meth_ptr);
      |                 ^~~~~~~~~~~~~~~~~~~~
      |                 EVP_CIPHER_CTX_free
```
...which comes from:
```c
void
cipher_free(struct sshcipher_ctx *cc)
{
[...]
#ifdef WITH_OPENSSL
        EVP_CIPHER_CTX_free(cc->evp);
        cc->evp = NULL;
        /* if meth_ptr isn't null then we are using the aes_ctr_mt
         * evp_cipher_meth_new() in cipher-ctr-mt.c under OSSL 1.1
         * if we don't explicitly free it then, even though we free
         * the ctx it is a part of it doesn't get freed. So...
         * cjr 2/7/2023
         */
        if (cc->meth_ptr != NULL) {
                EVP_CIPHER_meth_free((void *)(EVP_CIPHER *)cc->meth_ptr);
                cc->meth_ptr = NULL;
        }
#endif
[...]
}
```
HPN-SSH upstream also still has `EVP_CIPHER_meth_free()`, see https://github.com/rapier1/hpn-ssh/blob/master/cipher.c#L586 ATM.

So I guess, GSI-OpenSSH won't build with OpenSSL 4.x for now.

fscheiner:
@msalle , @ellert : I guess we can merge this anyhow, as in reality, that the build of the in-tree GSI-OpenSSH fails with OpenSSL 4.0 right now (e.g. only affects Debian Sid with _experimental_ suite) is not that big an issue, I'd say. This can surely be fixed later.

So, if you all agree, let's merge this.

@ellert : Do we need to make a new release so you can update the Debian packages or will it work without?

ellert:
> @ellert : Do we need to make a new release so you can update the Debian packages or will it work without?

It is not strictly necessary to have a release, since fixes can be applied as patches in packaging.

msalle (approved):
I think we've done sufficient testing for now. We know the gsi-ssh won't build yet, but that can wait till a later stage.

fscheiner (approved):
Looks good to me!

I wonder if all those _single use_ vars that caught the return value of a function where used to determine that value when running in a debugger or so. But e.g. `gdb` seems to also allow to determine that without an extra var - see https://stackoverflow.com/a/267687 - so not really needed, I guess.

fscheiner:
Merging that now.

@ellert: Many thanks for all the heavy-lifting you do. We weren't here without you, that's for sure!

ellert:
> So I guess, GSI-OpenSSH won't build with OpenSSL 4.x for now.

gsi-openssh 10.3p1 is in Fedora rawhide compiled with openssl 4.

